Issue #1, opened 2024-04-25
Create the empty documentation pages


arewm:
In order to enable easy parallelization of the documentation writing, we should create empty files associated with this outline:

```
Discover
  Why Konflux
  Supply chain security through SLSA conformity

Get Started
  Getting started with Konflux
  Getting started with CLI

How-tos 
  Creating applications and components	
  Configuring your builds 
    Customizing the build pipeline
    Creating secrets for your builds
    Configuration as code
    Enabling hermetic builds
    Prefetching package manager dependencies for hermetic build
    Defining component relationships
    Preventing redundant rebuilds
  Testing your application
    Overview of Konflux tests
    Surface-level tests
      Enabling a Synk task
    Adding an integration test
    Creating a custom integration test
    Retriggering integration tests
  Inspecting your supply chain metadata
    Inspecting SBOMs
    Downloading your SLSA provenance
  Deleting applications and components


Advanced How-tos
  Installing
  Managing a security fix
  Managing multiple software versions
  Managing compliance with the Enterprise Contract
  Releasing an application 
    Creating a release plan
    Creating a release plan admissions

Glossary
Contribute to documentation
```

PR #10, opened 2024-04-26
Initial doc layout with left navigation and page stubs

arewm:
Resolves: konflux-ci/docs#1


arewm:
Preview of the sidebar from a local render:
Screenshot 2024-04-26 at 10 18 03 AM

arewm:
I missed the expansion of a couple of sections in the previous screengrab.
Screenshot 2024-04-26 at 10 20 17 AM

TanyaShearonRH:
@arewm From comparison to what we mapped out, the above screen cap of navigation matches what we had in gdoc. I didn't see anything missing. 

Few comments/questions:
1) From one of the interviews of teams onboarding, the onboarding team's platform architect described needing to create the release plan admissions and reaching out to Konflux team. I"m assuming the onboarding team does not create the release plan object or release plan admissions---which is why we have releasing an application in the Advanced how-tos. Is this assumption correct? 

2) I would change title from "Installing" to "Installing Konflux" 

3) I wonder if we should re-order the topics in the Advanced How-tos section based on order in which that role would do complete the topics. For example,

Installing Konflux
Managing compliance with Enterprise Contract
Release an application
     Creating a release plan
     Creating a release plan admission
Managing multiple software versions
Managing a security fix

TanyaShearonRH:
@arewm Also, consider changing "Inspecting your supply chain metadata" container to "Inspecting provenance and attestations."

arewm:
@TanyaShearonRH, thanks for that feedback. Sounds good on all of it, see the latest commit (96abcdd) for the changes.

We can move the releases from advanced how-tos to the "normal" how-tos later if needed. The normal/advanced nature depends on how specific deployments of Konflux are configured and how they will be used. I left it in the advanced section for now.

ryandeussing:
LGTM!

brianwcook (approved):

PR #22, opened 2024-05-09
Adding SLSA conformity

missmesss:
Copying the SLSA conformity doc published for AppStudio.
The only update is the xref at the very end of the doc.

### DIFF
--- a/docs/modules/ROOT/pages/discover/slsa-conformity.adoc
+++ b/docs/modules/ROOT/pages/discover/slsa-conformity.adoc
@@ -4,7 +4,7 @@
 
 Supply-chain Levels for Software Artifacts (SLSA) is a security framework produced through industry collaboration. We use this framework as a guide for reinforcing the build process we use for your applications, to better secure your software supply chain.
 
-SLSA assigns two primary responsibilities to build platforms like {ProductName}:
+The SLSA Build (v1.0) specification assigns two primary responsibilities to build platforms like {ProductName}:
 
 * *Provenance* to describe how the platform built each software artifact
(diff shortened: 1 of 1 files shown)


missmesss:
@ralphbean Will you please confirm this doc is good to go for Konflux? The only change is an xref.

A comment from Andrew: I don't see anything apparent off hand other than the links. It is all still pretty generic and SLSA hasn't changed much since it was written either.

Thanks!

docs/modules/ROOT/pages/discover/slsa-conformity.adoc
```diff
@@ -1 +1,150 @@
 = Supply chain security through SLSA conformity
+
+== Supply-chain Levels for Software Artifacts (SLSA)
+
+Supply-chain Levels for Software Artifacts (SLSA) is a security framework produced through industry collaboration. We use this framework as a guide for reinforcing the build process we use for your applications, to better secure your software supply chain.
+
+SLSA assigns two primary responsibilities to build platforms like {ProductName}:
```
arewm, line 7:
```suggestion
The SLSA Build (v1.0) specification assigns two primary responsibilities to build platforms like {ProductName}:
```
The specification may change, so we can specifically indicate which track/version we are referencing: https://slsa.dev/spec-stages#versioning

This is also consistent with the additional links at the bottom.

arewm (changes requested):

missmesss:
Thank you for a review @arewm , I've added the spec version as you've suggested.

arewm (approved):

PR #29, opened 2024-05-14
Moved content for Retriggering integration tests for upstream

gtrivedi88:
Purpose: The purpose of this PR is to develop the releasing an application content. I have copied the AppStudio content directly. We have to review and update it.


docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
```diff
@@ -1 +1,57 @@
...
+= Retriggering Integration Tests
+
+Integration test scenarios for a given snapshot can be re-triggered by adding a label to the snapshot.
+
+.Prerequisites
+- You have a snapshot that has completed all its initial tests. All initial tests must be finished before you trigger subsequent tests.
+
+- The snapshot for which tests will be retriggered has completed its initial tests.
+
+- You have CLI access to the specific OpenShift cluster. For information on obtaining CLI access, refer Getting started.
```
arewm, line 10: Will we point readers there with a link/reference?

docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
```diff
@@ -1 +1,57 @@
...
+
+- You have an up-to-date kubectl binary.  Alternatively, the `oc` binary is also compatible.
```
arewm, line 12: This isn't _necessarily_ a requirement but I don't know what the version requirements are. This also feels like more pertinent information to have in the CLI access section than here.

docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
```diff
@@ -1 +1,57 @@
...
+
+.Procedure
+
+. Identify the IntegrationTestScenario that needs to be rerun.  For a given snapshot, you can only rerun one IntegrationTestScenario.
```
arewm, line 16: @dirgim @14rcole , do we need to provide assistance here to instruct users in ways to identify the desired Snapshot?

docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
```diff
@@ -1 +1,57 @@
...
+----
+$ kubectl label snapshot [snapshot name] test.appstudio.openshift.io/run=[scenario name]
+----
+
+. The tests are re-triggered automatically.  Once they are re-triggered, the system removes the label, allowing you to apply a new label for a different scenario if you wish to test multiple scenarios.
+
++
+[source]
+----
+$ tkn pipelinerun list
```
arewm, line 31: There is no context given to this command. Is this just to verify that the pipeline was kicked off? Should a caveat be given that no triggered pipelinerun may be found if too much time has elapsed? Or that there may be more pipelineruns returned if more work is happening in the cluster?

docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
dirgim: The related Snapshot can be found in the pipelineRun details page in the UI, e.g. 
image

docs/modules/ROOT/pages/how-tos/testing/integration/rerunning.adoc
gtrivedi88: @dirgim Please let us know your thoughts?

dirgim: Agreed, both of those caveats make sense to me, something like:
```
Note: this command is only to verify that a re-triggered integration pipelineRun is started. If it's run a long time after the re-trigger command, it may not show any output since pipelineRun resources are pruned after completion. Additionally, the command may show other, unrelated pipelineRuns that are in flight at that same time.
```

dirgim: @MartinBasti fyi ^^

MartinBasti (approved):

dirgim (approved):
+1

PR #34, opened 2024-05-16
Update index.adoc

Chr1st1anSears:
Starting off this draft for the parent page for testing

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/testing/index.adoc
+++ b/docs/modules/ROOT/pages/how-tos/testing/index.adoc
@@ -14,7 +14,7 @@ Build-time tests in {ProductName} ensure the stability and security of the appli
 
 The build-time tests used in {ProductName} are executed in the form of Tekton xref:glossary/index.adoc#task[tasks]. The utility used for validating container information is link:https://www.conftest.dev/[conftest]. A full listing of {ProductName} surface-level tests is available in this document: xref:./build/index.adoc[Build-time tests].
 
-Be aware that build-time tests include the following security-focused tests:
+Build-time tests include the following security-focused tests:
 
 * Vulnerability scanning via Clair
 * Anti-virus scanning via ClamAV
@@ -28,6 +28,6 @@ Custom tests in {ProductName} are tests that users and administrators create. To
 
 == Integration tests
 
-Integration tests ensure that all build components are able to work together at the same time. You can xref:./integration/adding.adoc[add an integration test], simply by giving {ProductName} the address to a GitHub repo, and the path within that repo to the `.yaml` file that defines the test.
+Integration tests ensure that all build components within an application are able to work together at the same time. You can xref:./integration/adding.adoc[add an integration test], simply by giving {ProductName} the address to a GitHub repo, and the path within that repo to the `.yaml` file that defines the test.
 
-{ProductName} runs integration tests after it successfully builds the components of an application. As part of the build process, {ProductName} creates an image for each component and stores them in a Quay.io repository. Images of all the components are then compiled into a snapshot of the application. {ProductName} tests the snapshot against user-defined IntegrationTestScenarios, which, again, refer to a GitHub repository. 
+{ProductName} runs integration tests after it successfully builds the variou components of an application. As part of the build process, {ProductName} creates an image for each component and stores them in a repository. Images of all the components are then compiled into a snapshot of the application. {ProductName} tests the snapshot against user-defined IntegrationTestScenarios, which, again, refer to a GitHub repository. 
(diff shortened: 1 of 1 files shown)


Chr1st1anSears:
The [guidance for this issue](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=19646&projectKey=HACDOCS&view=detail&selectedIssue=HACDOCS-923#) suggested splitting content between an overview page and the parent page for the whole testing section. I noticed that the doc for that parent page suggested that we might just put all the content on the parent page.

FWIW, from a writer's perspective, it makes most sense to just delete the overview page, and move everything to the parent page. So that's what I did here--upgrading the 'might' to a 'definitely,' for moving all content to the parent page. But LMK if you think we should do differently.

docs/modules/ROOT/pages/how-tos/testing/index.adoc
```diff
@@ -1 +1,33 @@
...
+
+== Custom tests
+
+Custom tests in {ProductName} are tests that users and administrators create. To add a custom test to an individual component, xref:/how-tos/configuring/customizing-the-build.adoc[customize its build pipeline] to include the test as another Tekton task. Or, to add a test that runs on all components of an application, xref:./integration/creating.adoc[create a custom integration test].
+
+== Integration tests
+
+Integration tests ensure that all build components within an application are able to work together at the same time. You can xref:./integration/adding.adoc[add an integration test], simply by giving {ProductName} the address to a GitHub repo, and the path within that repo to the `.yaml` file that defines the test.
+
+{ProductName} runs integration tests after it successfully builds the variou components of an application. As part of the build process, {ProductName} creates an image for each component and stores them in a repository. Images of all the components are then compiled into a snapshot of the application. {ProductName} tests the snapshot against user-defined IntegrationTestScenarios, which, again, refer to a GitHub repository.
```
arewm, line 33:
```suggestion
{ProductName} runs integration tests after it successfully builds the various components of an application. As part of the build process, {ProductName} creates an image for each component and stores them in a repository. Images of all the components are then compiled into a snapshot of the application. {ProductName} tests the snapshot against user-defined IntegrationTestScenarios, which, again, refer to a GitHub repository. 
```

docs/modules/ROOT/pages/how-tos/testing/index.adoc
```diff
@@ -1 +1,33 @@
...
+
+* Vulnerability scanning via Clair
+* Anti-virus scanning via ClamAV
+* Code scanning via SAST tools
++
+** You can enable the Snyk test by following xref:./build/snyk.adoc[these instructions].
+
+== Custom tests
+
+Custom tests in {ProductName} are tests that users and administrators create. To add a custom test to an individual component, xref:/how-tos/configuring/customizing-the-build.adoc[customize its build pipeline] to include the test as another Tekton task. Or, to add a test that runs on all components of an application, xref:./integration/creating.adoc[create a custom integration test].
```
arewm, line 27: platform engineers might be better than administrators.

docs/modules/ROOT/pages/how-tos/testing/index.adoc
```diff
@@ -1 +1,33 @@
...
+
+Build-time tests include the following security-focused tests:
+
+* Vulnerability scanning via Clair
+* Anti-virus scanning via ClamAV
+* Code scanning via SAST tools
++
+** You can enable the Snyk test by following xref:./build/snyk.adoc[these instructions].
+
+== Custom tests
```
arewm, line 25: These are custom build-time tests specifically. Not custom IntegrationTestScenarios. Do we want to add anything about the latter?

arewm (approved):
Just nits, no additional review needed.

docs/modules/ROOT/pages/how-tos/testing/index.adoc
Chr1st1anSears: Oh, hmm, I thought that the second type of custom tests mentioned in this section would be a custom ITS. Because you create a YAML file and it defines a test that runs on the whole snapshot. Is that not quite right?

I'll at least edit the second sentence, while we're still talking about custom build-time tests, to say, "To add a custom build-time test..." Hopefully that gives the doc enough clarity.

Chr1st1anSears: Given what you said about the review in general, I'm gonna hit the green button. But if you'd like me to adjust anything, I'll be happy to reopen it and make any corrections we need.

arewm: We can come back and split up that paragraph later. It isn't quite clear with a second reading. You are talking about two different types of custom tests in the same paragraph. The first is a built time test and the second is a custom ITS.

It might make sense to have two sections -- build-time and integration tests. In each section, we can provide directives for how to customize.

PR #37, opened 2024-05-17
Update component-nudges.adoc

Chr1st1anSears:
Starting this draft with existing content from previous doc set

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
@@ -6,6 +6,12 @@ In such instances, whenever you build a new image of the common parent component
 
 However, if you build an application with {ProductName} and make references between components with image digests (`sha256:...`), {ProductName} can automatically generate pull requests to update those references. To use this functionality, simply define the relationships between your components, either in the {ProductName} UI, or in your CLI.
 
+[NOTE]
+====
+By default, {ProductName} only looks for image digest references in certain types of files. However, you can xref:customizing-the-build.adoc[customize your build pipeline] to adjust that list of files with an annotation. Add the following annotation under `annotations` in your component's `/.tekton/<component-name>-push.yaml` file:
+`build.appstudio.openshift.io/build-nudge-files: "<regex for file type, like .*.md>"`
+====
+
 == In the UI
 
 .Procedure
--- a/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
@@ -8,7 +8,7 @@ However, if you build an application with {ProductName} and make references betw
 
 [NOTE]
 ====
-By default, {ProductName} only looks for image digest references in certain types of files. However, you can xref:customizing-the-build.adoc[customize your build pipeline] to adjust that list of files with an annotation. Add the following annotation under `annotations` in your component's `/.tekton/<component-name>-push.yaml` file:
+By default, {ProductName} only looks for image digest references in certain types of files. However, you can xref:./customizing-the-build.adoc[customize your build pipeline] to adjust that list of files with an annotation. Add the following annotation under `annotations` in your component's `/.tekton/<component-name>-push.yaml` file:
 `build.appstudio.openshift.io/build-nudge-files: "<regex for file type, like .*.md>"`
 ====
 
--- a/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
@@ -1,1 +1,44 @@
 = Defining component relationships
+
+When an application includes multiple components, their repositories might contain references to image builds of another component. For instance, an application might include multiple child components that, using a Dockerfile `FROM` clause, reference the image build of a common parent.
+
+In such instances, whenever you build a new image of the common parent component, you need to update the references to it. Specifically, you need to copy and paste the pullspec and image digest of that new image to your other source repositories. This process is tedious and error-prone.
+
+If you build an application with {ProductName} and make references between components with image digests (`sha256:...`), {ProductName} can automatically generate pull requests to update those references. To use this functionality, simply define the relationships between your components in the {ProductName} UI.
+
+.Procedure
+
+To define the relationships between components in an application, complete the following steps in the {ProductName} UI:
+
+. Navigate to your application. 
+. On any tab, use the *Actions* drop-down menu to select *Define component relationships*. 
+.. Alternatively, go to the *Components* tab and select the *Define component relationships*.
+. In the *Component relationships* window, select one component from the *Select a component* drop-down menu.
+. Select *Nudges* or *Is nudged by*, depending on the relationship you wish to define.
+
++
+[NOTE]
+====
+Component A nudges Component B (or Component B is nudged by Component A) if Component B contains a reference by image digest to a build of Component A. 
+====
+
+. Use the remaining drop-down menu to choose which other components belong to this relationship.
+. To define multiple relationships, select *Add another component relationship*.
+. Once you have defined all necessary relationships, select *Save relationships*.
+
+.Verification
+
+To verify a relationship between your components, complete the following steps:
+
+. Go to the *Components* tab for your application. 
+. Select a component that belongs to the relationship you defined.
+. Scroll to the end of the page and select *Show related components*. 
+. Merge a pull request for Component A. When the build completes, you should see a new pull request appear for Component B. This new pull request contains the  image digest for the new build of Component A.
+
+
+.Troubleshooting
+
+To resolve any issues with relationships between components, complete the following steps:
+
+. On any tab, use the *Actions* drop-down menu to select *Define component relationships*. If you do not see the relationship, try to define it again and then make sure to select *Save relationships*.
+. Ensure that the existing references in the repositories of your components are correct. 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -1 +1,81 @@
...
+. Navigate to your application. 
+. On any tab, use the *Actions* drop-down menu to select *Define component relationships*. 
+.. Alternatively, go to the *Components* tab and select the *Define component relationships*.
+. In the *Component relationships* window, select one component from the *Select a component* drop-down menu.
+. Select *Nudges* or *Is nudged by*, depending on the relationship you wish to define.
+
++
+[NOTE]
+====
+Component A _nudges_ Component B (or Component B _is nudged by_ Component A) if Component B contains a reference by image digest to a build of Component A.
```
arewm, line 24: and that digest should be updated after successful push builds of Component A

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -1 +1,81 @@
...
+. Select a component that belongs to the relationship you defined.
+. Scroll to the end of the page and select *Show related components*. 
+
+Alternatively, in the git repo for the component that nudges, push a commit. When the build for that component completes, you should see a new pull request appear for the component that is nudged. This new pull request contains the image digest for the new build of its parent image.
+
+
+.Troubleshooting
+
+* On any tab, use the *Actions* drop-down menu to select *Define component relationships*. If you do not see the relationship, try to define it again and then make sure to select *Save relationships*.
+* Ensure that the existing references in the repositories of your components are correct.
```
arewm, line 44: It is also possible to troubleshoot the _nudging_. This will happen via a Renovate pipeline run in the tenant workspace (you will have to use the CLI to see the PLR as it will not show up in the UI).

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -1 +1,81 @@
-= Defining component relationships
\ No newline at end of file
+= Defining component relationships
+
+When an application includes multiple components, their repositories might contain references to image builds of another component. For instance, an application might include multiple child components that, using a Dockerfile `FROM` clause, reference the image build of a common parent.
+
+In such instances, whenever you build a new image of the common parent component, you need to update the references to it. Specifically, you need to copy and paste the pullspec and image digest of that new image to your other source repositories. This process is tedious and error-prone.
+
+If you build an application with {ProductName} and make references between components with image digests (`sha256:...`), {ProductName} can automatically generate pull requests to update those references. To use this functionality, simply define the relationships between your components, either in the {ProductName} UI, or in your CLI.
```
arewm, line 7: By default, it only looks in certain files for these references: https://github.com/konflux-ci/build-service/blob/0e591f628aab9cbe9fac2b9bdc8883e955a560e7/controllers/component_dependency_update_controller.go#L287C18-L287C60 ... which can be customized by setting an Annotation on the pipeline definition: https://github.com/konflux-ci/build-service/blob/0e591f628aab9cbe9fac2b9bdc8883e955a560e7/controllers/component_dependency_update_controller.go#L65

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
Chr1st1anSears: I think I follow what you're saying. For example, users could delete '.*.yaml' if, for some reason, they don't want Konflux to search for references in yaml files, right? Or add another file type to look for.

But how would users change that definition, given that it's in the build service? They only clone and use this repo: https://github.com/konflux-ci/konflux-ci
Is there a file in that repo with the same annotation to edit?

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
Chr1st1anSears: I'm afraid you lost be a bit here--exactly what steps would users take to troubleshoot the nudging in the CLI?

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
arewm: The would just add an annotation to their PipelineDefinition, i.e. in here: https://github.com/spork-madness/common-nodejs-parent/blob/main/.tekton/common-nodejs-parent-push.yaml#L4

This is more likely to be used if you want to nudge references in different files.

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
arewm: `oc get pr` i.e. the same flow as using the `oc` CLI to get information from the workspace, just getting pipelineruns

Chr1st1anSears: Ok so you use `oc` to get the PR itself. Makes sense. But then users would need to do something with that PR, too, right?  It sounded like something involving Renovate Bot?

arewm: It might be too much detail for here. Maybe better for a SOP? They would need to get the logs for the pipeline run to see if it is working properly or if it is amiss.

Chr1st1anSears: Ok so should I just leave it out for this current iteration of the doc? Something we come back to?

arewm: Yeah, I talked myself into leaving it out.

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
Chr1st1anSears: I took a stab at covering this in a note. Please take a look and see what you think.

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -1 +1,87 @@
...
+
+In such instances, whenever you build a new image of the common parent component, you need to update the references to it. Specifically, you need to copy and paste the pullspec and image digest of that new image to your other source repositories. This process is tedious and error-prone.
+
+However, if you build an application with {ProductName} and make references between components with image digests (`sha256:...`), {ProductName} can automatically generate pull requests to update those references. To use this functionality, simply define the relationships between your components, either in the {ProductName} UI, or in your CLI.
+
+[NOTE]
+====
+By default, {ProductName} only looks for image digest references in Dockerfiles, Containerfiles, and YAML files. However, you can xref:./customizing-the-build.adoc[customize your build pipeline] to adjust that list of files with an annotation. Add the following annotation under `annotations` in your component's `/.tekton/<component-name>-push.yaml` file:
```
arewm, line 11: Nit: it is probably fine just to point them to the push file. We don't support nudging on PRs for now.

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -1 +1,87 @@
...
+`build.appstudio.openshift.io/build-nudge-files: "<regex for file type, like .*.md>"`
```
arewm, line 12: I think that these can technically be regex or glob patterns, but it's fine as it is: https://docs.renovatebot.com/string-pattern-matching/ ... I think we can also say a comma separated list of regexes?

Do you want to put the _default_ regex in here so that users can see how the pattern might look?

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
Chr1st1anSears: I think I already point them to the push file, since I say "Add the following annotation...in your component's...*push.yaml* file"
Or am I not following what you meant?

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
Chr1st1anSears: I revised the note to include glob patterns and provide the list of existing regex patterns as examples. Please LMK if we're good to go, or need to add anything else.

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
arewm: I don't know what I was thinking there. But now I am thinking -- can we just explicitly state that this functionality is only enabled for push builds?

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
arewm: Did you push the revision? I don't see a change.

Chr1st1anSears: Yeah I pushed the revision; I can see it here: https://github.com/konflux-ci/docs/blob/Chr1st1anSears-patch-6/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc

Chr1st1anSears: If you look at lines 11-14 (the whole Note) I think they now address your original point about directing where Konflux looks for image references

PR #38, opened 2024-05-21
Updated build time tests

gtrivedi88:
Purpose: The purpose of this PR is to review and update build time tests content. I have copied and updated the AppStudio content. Next tasks is to review and update it.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
+++ b/docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
@@ -1,1 +1,51 @@
 = Build-time tests
+
+This document covers the build-time tests that {ProductName} runs as part of its component build pipeline. These build-time tests automatically check all application images to inform you if they're up-to-date, correctly formatted, and protected from security vulnerabilities.
+
+NOTE: These tests are non-blocking and do not prevent the pipeline from proceeding if issues are found.
+
+The component build pipeline for {ProductName} include various types of build-time tests, conducted using link:https://tekton.dev/docs/pipelines/tasks/#overview[Tekton tasks].
+
+The table below outlines the default build-time tests:
+
+.Deprecated image checks
+|===
+|Test name |Description |Failure message
+
+|image_repository_deprecated |Deprecated images are no longer maintained, leading to unresolved security vulnerabilities. | The container image must not be built from a repository  marked as 'Deprecated' in COMET
+|===
+
+.Unsigned RPM check
+|===
+|Test name |Description |Failure message
+
+|image_unsigned_rpms |Packages signed with Red Hat's secure signing server adheres to stringent policies and procedures. |Alerts the user for any unsigned RPMs. Found following unsigned rpms(nvra):
+|===
+
+.Security checks
+|===
+|Test name |Description |Failure message
+
+|clair-scan |Scans container images for vulnerabilities using Clair, by comparing the components of container image against Clair's vulnerability databases. | Found packages with critical vulnerabilities. Consider updating to a newer version of those packages, they may no longer be affected by the reported CVEs.
+
+|clamav-scan |Scans the content of container images for viruses, malware, and other malicious content using ClamAV antivirus scanner. | A malware has been found.
+
+|sast-snyk-check |Scans and analyze source code or compiled versions of code to help find security flaws using Snyk Code. | For details, review Tekton task log.
+|===
+
+You can disable the default build-time tests by setting the `skip-checks` parameter to `true`
+
+[source,yaml]
+----
+- description: Skip checks against built image
+  name: skip-checks
+  type: string
+  default: "false"
+----
+ 
+[role="_additional-resources"]
+== Additional resources
+
+* For information about enabling Snyk task, see xref:how-tos/testing/build/snyk.adoc[Enabling a Snyk task].
+
+* For additional build-time tests, see link:https://konflux-ci.dev/docs/how-tos/configuring/customizing-the-build/[Customizing the build pipeline].
--- a/docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
+++ b/docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
@@ -1,6 +1,54 @@
-= Enabling Snyk
+= Enabling a Snyk task
 
-While a Snky test is available to run a build time in the default pipelines, additional configuration is needed to enable the test. This is an example of a build-time test that requires the configuration of a custom secret to be able to run.
+The Snyk test is available to run at build time in the default pipelines, but it requires additional configuration to enable. This procedural example illustrates a build-time test that requires the configuration of a custom secret.
 
-+
-NOTE:  We may want to consider generalizing this topic.
+The `sast-snyk-check` task uses the Snyk Code tool to perform static application security testing (SAST). 
+Specifically, the Snyk check scans an application's source code for potential security vulnerabilities, 
+including SQL injection, cross-site scripting (XSS), and code injection attack vulnerabilities.
+
+NOTE: You can run a Snyk task only if you have a Snyk token configured in a namespace secret. Ensure that the name of your secret is included in the *snyk-secret* pipeline parameter.
+
+.Procedure
+
+. Register for a Snyk account or log in at https://app.snyk.io/.
+. Get a Snyk token.
+
+.. In the lower left of the home page, click your name, then select *Account settings*.
+
+.. From the Account Settings page, select *General*, which is the default, then *Auth Token*.
+
+.. Under the *Auth Token* section, click *Click to View* to see the *KEY* value of the automatically generated token.
+
+. Enable Snyk Code.
+
+.. From the left panel, go to *Settings* > *Snyk Code*, then scroll to the *Enable Snyk Code* section.
+
+.. Toggle *Disabled* to *Enabled*.
+
+.. Click *Save* changes.
+
+. Add your new secret to your workspace.
+
+.. Log in to {ProductName} Overview page.
+
+.. From the left menu, click *Secrets*.
+
+.. Click *Add secret*.
+
+.. The *Add secret* page displays options for your new secret. Specify the following:
+
+... For *Secret for*, select *Build*.
+
+... From the *Secret type* drop-down menu, choose *Key/value secret*.
+
+... From the *Secret name* drop-down menu, select *snyk-secret*.
+
+... Paste your Snyk token into the *Upload the file with value for your key or paste its contents* field.
+... Click *Add secret* to save it.
+
+You've enabled the Snyk task for your build pipeline.
+
+[role="_additional-resources"]
+== Additional resources
+
+For more information about Snyk, see link:https://snyk.io/product/snyk-code/[the Snyk website].


docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
-= Build-time tests
\ No newline at end of file
+= Build-time tests
+
+This document covers the build-time tests (formerly known as "sanity tests") that {ProductName} runs as part of its component build pipeline. These build-time tests automatically check all application images to ensure that they're up-to-date, correctly formatted, and protected from security vulnerabilities.
```
arewm, line 3: I don't think we need to call out the former name.

arewm, line 3: Nit: the out-of-the-box tests don't "ensure" as much as they "inform" since the tasks are non-blocking/failing.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
...
+
+The component build pipeline for {ProductName} include various types of build-time tests, conducted using link:https://tekton.dev/docs/pipelines/tasks/#overview[Tekton tasks]. For validating container information, the pipeline employs https://www.conftest.dev/[Conftest]. The table below outlines the currently implemented build-time tests:
```
arewm, line 5: The tests that are included in the pipeline by default instead of currently implemented?

arewm, line 5: @dirgim , in your opinion, is the use of Conftest a relevant implementation detail to reveal to users or should we remove that bit?

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
...
+
+.Deprecated image checks
+|===
+|Test name |Description |Failure message
+
+|image_repository_deprecated |Deprecated images are no longer maintained, leading to unresolved security vulnerabilities. | The container image must not be built from a repository  marked as 'Deprecated' in COMET
```
arewm, line 11: COMET is likely irrelevant for external users. Can we change this?
@dirgim, should the task itself be changes as it is exposed to the community? I assume that this is only relevant for Red Hat image references.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
...
+|===
+
+.Unsigned RPM check
+|===
+|Test name |Description |Failure message
+
+|image_unsigned_rpms |Packages signed with Red Hat's secure signing server adheres to stringent policies and procedures. |All RPMs in the image must be signed. Found following unsigned rpms(nvra):
```
arewm, line 18: The test doesn't enforce so it is hard to say something like "all RPMs in the image must be signed" instead, it may be better to say that it alerts if there are any unsigned RPMs included.

Does it make sense to change the first sentence as well to add the relevant perspective for this use case without assuming that users are within Red Hat?

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
...
+|===
+
+
+[role="_additional-resources"]
```
arewm, line 22: What are roles and what does this do out of curiosity?

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
```diff
@@ -1 +1,27 @@
...
+== Additional resources
+
+* Snyk test is also available to run at build time in the default pipelines.For more information about enabling Snyk task, see xref:how-tos/testing/build/snyk.adoc[Enabling a Snyk task].
```
arewm, line 25: This task is only special by how it is enabled. I think it can still go into the table above.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
arewm, line 5: We are missing some tests here like clair, clamav, and the ecosystem preflight check.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
arewm, line 3: These tests are all enabled by default by configuration of the default parameter:
```
    - description: Skip checks against built image
      name: skip-checks
      type: string
      default: "false"
```
Do we want to inform users how to disable them by setting that parameter?

docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
```diff
@@ -1,6 +1,54 @@
...
+The Snyk test is available to run at build time in the default pipelines, but it requires additional configuration to enable. This procedural example illustrates a build-time test that requires the configuration of a custom secret.
 
-+
-NOTE:  We may want to consider generalizing this topic.
\ No newline at end of file
+The `sast-snyk-check` task uses the Snyk Code tool to perform static application security testing (SAST). 
+Specifically, the Snyk check scans an application's source code for potential security vulnerabilities, 
+including SQL injection, cross-site scripting (XSS), and code injection attack vulnerabilities.
+
+NOTE: You can run a Snyk task only if you have a Snyk token stored in a namespace secret. Ensure that the name of your secret is included in the *snyk-secret* pipeline parameter.
```
arewm, line 9: Configured instead of stored?

docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
```diff
@@ -1,6 +1,54 @@
...
+.Procedure
+
+. Register for a Snyk account or log in at https://app.snyk.io/.
+. Get a Snyk token.
+
+.. In the lower left of the home page, click your name, then select *Account settings*.
+
+.. From the Account Settings page, select *General*, which is the default, then *Auth Token*.
+
+.. Under the *Auth Token* section, click *Click to View* to see the *KEY* value of the automatically generated token.
```
arewm, line 20: Is there Snyk documentation that we can/should point to? I assume that this is relatively stable, but in case it isn't do we want to keep it updated?

Having the documentation all in one place is beneficial ... so I am just asking the question, not requesting a change.

docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
```diff
@@ -1,6 +1,54 @@
...
+
+.. From the left panel, go to *Settings* > *Snyk Code*, then scroll to the *Enable Snyk Code* section.
+
+.. Toggle *Disabled* to *Enabled*.
+
+.. Click *Save* changes.
+
+. Add your new secret to your workspace.
+
+.. Log in to the link:https://console.redhat.com/preview/application-pipeline[Red Hat Hybrid Cloud Console {ProductName} Overview page].
```
arewm, line 32: This link isn't relevant anymore.

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
gtrivedi88: Actually, it's from our doc template. If we use this format, the system formats the output in a specific way.

docs/modules/ROOT/pages/how-tos/testing/build/snyk.adoc
gtrivedi88: We have the link to Snyk docs in Additional resources. Let me know if that works.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
gtrivedi88: @arewm How about this?

`You can disable the default build-time tests by setting the `skip-checks` parameter to `true`

[source,yaml]
----
- description: Skip checks against built image
  name: skip-checks
  type: string
  default: "false"
----`

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
gtrivedi88: @arewm So do we want to add tables for each of these, like we have now?

- Vulnerability scanning via Clair
- Anti-virus scanning via ClamAV
- Code scanning via SAST tools

OR

Do we want to update the existing paragraph with these. For example: The pipeline employs:

- link:https://www.conftest.dev/[Conftest] to Validateg container information 
- Clair to scan vulnerabilities in the code.
- ClamAV for anti-virus scan
- Static Application Security Testing (SAST) tools to scan and analyze source code or compiled versions of code to help find security flaws

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
dirgim: That's correct, the [konflux-test policy](https://github.com/konflux-ci/konflux-test/blob/9bc1111b160a215bedb4bc3cc9d3f866e9b52ff0/policies/repository/deprecated-image.rego#L7-L8) for this has already been updated to refer to the Red Hat Catalog instead of COMET.
My proposal is to use the same text here :
```
|image_repository_deprecated |Deprecated images are no longer maintained and will accumulate security vulnerabilities without releasing a fixed version. | The container image shouldn't be built from a repository that is marked as 'Deprecated' in Red Hat Catalog.
```

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
dirgim: I'm fine with removing that bit since the use of Conftest is abstracted away from the users.

dirgim: The integration team has the [STONEINTG-766](https://issues.redhat.com/browse/STONEINTG-766) ticket to add the test information for the security checks, I think that would cover most of them, apart from the ecosystem preflight check.

dirgim: @dheerajodha @MartinBasti  fyi ^^

dheerajodha: Thanks for tagging me!
Yes, I plan to open a PR soon, to add details of the remaining security checks here.

gtrivedi88: PR which addresses this comment - https://github.com/konflux-ci/docs/pull/47/files

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
gtrivedi88: Not sure if I get this, but based on my understanding I believe we can add a page like "Updating build time tests" and add a link here. WDYT?

arewm: It wouldn't be specific to build time tests as it is just about customizing the build pipeline (which potentially includes these build-time tests).

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
arewm: seems reasonable for now.

docs/modules/ROOT/pages/how-tos/testing/build/index.adoc
gtrivedi88: As discussed this in chat, I'll update this with content from other PR and will ask for your review. 
Marking this resolved.

gtrivedi88:
@arewm Added all the new checks from PR # 47. Could you please review and approve this PR.
@dirgim This is just FYI for the latest updates made on this PR.

dheerajodha (approved):
Thanks!

PR #39, opened 2024-05-21
Updated Deleting application

gtrivedi88:
Purpose: The purpose of this PR is to review and update Deleting application and component content. I have copied and updated the AppStudio content. Next tasks is to review and update it.


docs/modules/ROOT/pages/how-tos/deleting.adoc
```diff
@@ -1 +1,41 @@
-= Deleting applications and components
\ No newline at end of file
+= Deleting applications and components
+
+You can delete an application if you are the owner or if you have admin access to the application.
```
arewm, line 3: I would not say owner or admin access. Instead, you can say something like if you are a user with the workspace admin role.

docs/modules/ROOT/pages/how-tos/deleting.adoc
```diff
@@ -1 +1,41 @@
...
+
+WARNING: If you delete an application permanently, you delete all the components and resources from the user interface (UI) and upstream repositories. You cannot undo the deletion.
```
arewm, line 5: What resources are contained in upstream repositories? I guess that the deletion submits a PR to delete the tekton pipleine definitions. Is that what you were referring to?

docs/modules/ROOT/pages/how-tos/deleting.adoc
```diff
@@ -1 +1,41 @@
...
+
+== Deleting an application in the web UI
+.*Prerequisites*
+
+* You have successfully signed into {ProductName}.
+* You have at least one application.
```
arewm, line 11: These feel like no-op prerequisites. A better one might just be having the `admin` role.

docs/modules/ROOT/pages/how-tos/deleting.adoc
```diff
@@ -1 +1,41 @@
...
+
+.*Procedures*
+
+. In the console, select *My applications*.
```
arewm, line 15: Are we calling it a console or a UI?

docs/modules/ROOT/pages/how-tos/deleting.adoc
```diff
@@ -1 +1,41 @@
...
++
+NOTE: You can also delete an application by selecting *Actions > Delete Application*.
+
+
+== Deleting an application using CLI
+
+.*Prerequisites*
+
+* You have successfully signed into {ProductName}.
+* You have at least one application.
```
arewm, line 28: Similar content to above, but maybe also that you have configured the CLI?

gtrivedi88:
@arewm Addressed all comments. Could you please review and me know  your approval or suggestions?

PR #42, opened 2024-05-21
Add downloading slsa doc

Chr1st1anSears:


Chr1st1anSears:
Please check the very last bullet point, under "Additional resources." I believe this might address the concern you mentioned around EC in the Jira issue: https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=18221&projectKey=RHTAPBUGS&view=detail&selectedIssue=HACDOCS-929#

docs/modules/ROOT/pages/how-tos/metadata/provenance.adoc
```diff
@@ -1,4 +1,142 @@
-= Downloading the SLSA provenance
+= Downloading your SLSA provenance
```
arewm, line 1: Why this change? It is the provenance for some specific artifact, not necessarily related to the user

docs/modules/ROOT/pages/how-tos/metadata/provenance.adoc
```diff
@@ -1,4 +1,142 @@
...
 
+To deliver heightened build-time security, {ProductName} conforms to the requirements for link:https://slsa.dev/spec/v1.0/levels#build-l3[SLSA Build Level 3]. One of the most basic requirements is that we provide link:https://slsa.dev/spec/v1.0/provenance[provenance], which tells you the steps that {ProductName} takes when building your applications.
```
arewm, line 3: The phrase "to deliver heightened build-time security" feels odd but I don't have a different thought off hand.

Chr1st1anSears: Would "provide" or "support" be better than "deliver"?

arewm: Maybe part of the reason is that we didn't do it _because_ of SLSA. Conforming to SLSA is a side effect of our decisions.

Maybe something like this?
> With security intentionally designed into its core, {ProductName} happens to conform to the SLSA v1.0 Build Level 3 requirements. This includes providing transparency into the steps taken for all artifact builds via the generation of SLSA Provenance.

I am not committed to the words, but something along those lines?

Chr1st1anSears: So I def understand that SLSA isn't our "why," but I think "*happens* to conform" is also downplaying our commitment to those standards a little bit.

Here's my suggestion: "We are committed to providing exceptional build-time security with {ProductName}. As part of that commitment, {ProductName} conforms to..." and then I'd include everything you wrote, starting with "...the SLSA v1.0 Build Level 3 requirements."

Whaddaya think?

arewm: That sounds good, thanks!

PR #45, opened 2024-05-23
Update sboms.adoc

Chr1st1anSears:


docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
\ No newline at end of file
+= Inspecting SBOMs
+
+== Software bill of materials
+A software bill of materials (SBOM) provides greater transparency for your software supply chain. {ProductName} provides link:https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf[build SBOMs], which list all the software libraries that a component uses. Those libraries can enable specific functionality or facilitate development. 
+
+You can use an SBOM to better understand the composition of your software. In particular, you can analyze your SBOM to determine if your application has any vulnerable dependencies. Furthermore, to comply with cybersecurity regulations, you might need to provide your customers with the SBOM for your application's components. For example, in the United States, the link:https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/[Executive Order on Improving the Nation's Cybersecurity] requires vendors to provide SBOMs, and in Europe, the link:https://www.cisa.gov/sites/default/files/2023-09/EU%20Commission%20SBOM%20Work_508c.pdf[Cyber Resilience Act] does the same.
```
arewm, line 6: > you can analyze your SBOM to determine if your application has any vulnerable dependencies

The SBOM doesn't have any vulnerability information. In order to determine this, you would have to collate the data in the SBOM with some other data source (i.e. a vulnerability database).

arewm, line 6: > Furthermore, to comply with cybersecurity regulations, you might need to provide your customers with the SBOM for your application's components

It feels like the wrong abstraction to discuss customers here as users of the Konflux project would likely not have customers. Organizations that use the Konflux codebase might have customers... 

SBOMs are important to increase transparency into the supply chain, but that doesn't mean that everyone will start to produce them. Maybe we can just link to these types of regulations to highlight the additional driver/need to produce them and not talk to specifics about needing to provide SBOMs?

arewm, line 6: > does the same

I haven't read the CISA document. Does it actually do the same or just similar?

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
+
+== Viewing an SBOM in the web UI
+
+.Prerequisites
+
+* You must have an application that {ProductName} has successfully built and deployed.
```
arewm, line 12: A component is what needs to be built

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
+
+. For the component whose SBOM you want to view, select its most recent pipeline run.
+
+. On the *Pipeline run details* page, select *View SBOM*. 
+.. You can use your web browser to immediately search the SBOM for terms that indicate vulnerabilities in your software supply chain. For example, try searching for "log4j".
+.. You can select *Download* to download the SBOM, or *Expand* to view it full-screen.
+
+[NOTE]
+====
+This SBOM is unsigned. You can still use it for standard business processes, like providing an SBOM to your customers. However, do not use this unsigned SBOM as a legal document.
```
arewm, line 28: I would just generalize this and not give any specific use cases for it. Maybe something like:

The SBOM is unsigned so there is no way to verify whether it has been tampered with after being pushed to the registry. Additional integrity verification should be performed before distributing the artifact if needed.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
+----
+
+. Use Cosign to download the SBOM. From the output of the last command, pass the image path as an argument into Cosign's `download sbom` command. Be sure to delete any quotation marks around the image path.
+
++
+.Example
++
+[source]
+----
+$ cosign download sbom quay.io/redhat-appstudio/user-workload@sha256:<output omitted>
```
arewm, line 86: Maybe instead of `<output omitted>`, you can put `<sha-digest>` here, above, and below? It feels odd to have output labeled in the input.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
+----
+
++
+[NOTE]
+====
+Because  the SBOM is unsigned, Cosign displays a warning about its authenticity. You can still use this SBOM for standard business processes, like providing an SBOM to your customers. However, do not use this unsigned SBOM as a legal document.
```
arewm, line 92: I don't know if we need to repeat this note.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -1 +1,152 @@
...
+====
++
+[source]
+----  
+WARNING: Downloading SBOMs this way does not ensure its authenticity. 
+If you want to ensure a tamper-proof SBOM, download it using 'cosign download attestation <image uri>' 
+or verify its signature using 'cosign verify --key <key path> --attachment sbom <image uri>'.
```
arewm, line 99: Neither one of these options are valid in Konflux today so I don't know if it makes sense to include them.

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
Chr1st1anSears: That's exactly what I'm trying to communicate with "analyze your SBOM." For example, with RHTAP, we say you "analyze your SBOM with TPA" to get actionable info about vulns. 
Would this work?
 "In particular, you can use various tools to scan your SBOM and determine if any of your dependencies have known vulnerabilities."

Chr1st1anSears: What about just "...your *organization* might need to provide *its* customers..."

If we expect enterprises to use Konflux, then certainly an appeal of SBOMs from Konflux is for those enterprises to provide them to customers, to satisfy regulations.

Chr1st1anSears: Yep, it's almost identical language.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
Chr1st1anSears: Yeah so that's output that I got from cosign; it's not our own doc's commentary, if that makes sense. I can delete it, but I think it helps to show what users can expect to see after running the command.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
arewm: that seems reasonable.

arewm: you can simplify it further that they may need to be provided to consumers.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
```diff
@@ -9,7 +9,7 @@ You can use an SBOM to better understand the composition of your software. In pa
 
 .Prerequisites
 
-* You must have an application that {ProductName} has successfully built and deployed.
+* You must have a component that {ProductName} has successfully built and tested.
```
arewm, line 12: Nit: The ITS don't need to run, just the build pipeline. So these would just be the build-time tests (which by default should not cause the PipelineRun to fail even if the tests return issues).

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
arewm: I think it would just lead to confusion since it isn't currently relevant.

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
Chr1st1anSears: One thing ibm/rh hit us writers over the head with is that we should avoid the passive tense. Unless you've got a strong objection to it, I'm gonna use that wording: "your organization might need to provide its customers."

docs/modules/ROOT/pages/how-tos/metadata/sboms.adoc
Chr1st1anSears: Ok, I'll take it out

PR #51, opened 2024-05-30
HACDOCS-941-glossary

Chr1st1anSears:


Chr1st1anSears:
@arewm I didn't find any instances of Devfile. I deleted the one mention of GitOps. And I deleted all references to environments, except for one mention of production environment under managed workspace, because I thought that might still be valid. Please take a look at that definition and let me know if I should delete that. And of course I'm happy to make any other changes you'd like to see.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+**security testing** +
+A process that determines if images meet security quality standards.
+
+**signature and provenance** +
+Mechanisms that use Tekton chains to extract and store the metadata of the PipelineRuns, sign that metadata, and then store it in the image registry that is next to the component image.
+
+**snapshot** +
+A set of component and container images that specifies which components the system should release with which container images. The system creates a snapshot when it finishes running a component's build pipeline. 
+
+
```
arewm, line 78: extra line break

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+**integration test** +
+A pipeline that you set up in GitHub. When you add components, it tests each one individually, and then tests the application as a whole.
+
+**IntegrationTestScenario** +
+A parameter that the integration service uses to test application components.
+
+**Java build service** +
+A group of utilities that accelerates build times, locates dependencies in final images, and meets product security requirements by managing the construction of Java dependencies.
```
arewm, line 34: This isn't needed currently.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A platform to automate the process of building, testing, and releasing applications. {ProductName} offers enterprise-grade security and customizable feature sets.   
+
+**build pipeline customization** +
+The ability to update and manage build pipelines for each component in an application. 
+
+**cluster** +
+A Kubernetes deployment with nodes that run containerized applications and a control plane that manages the nodes.
+
+**commit** +
+A change to one or more files. In {ProductName}, commits that you make to a linked git repository move through the pipeline and automatically get published.
```
arewm, line 15: This is specific to content tracked in source control. These are not just any changes, but just those that have been added to the tree. 

https://git-scm.com/docs/git-commit
https://www.atlassian.com/git/tutorials/saving-changes/git-commit

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**component** +
+An image that {ProductName} builds from source code in a repository. One or more components that run together form an application.
```
arewm, line 18: It isn't necessarily an image. A Component is an internal representation of an artifact that is being built ... including the linking between the source git repository and the storage location in the OCI registry.

We are also not trying to necessarily say that they run together to form an application (we also don't define an application here).

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**Conftest** +
+A utility for testing structured configuration data. Use Conftest to validate container information.
```
arewm, line 21: Do we need to have this topic here? We can also refer to https://www.conftest.dev/ if needed ... or just refer to it where it is used to prevent it from needing to be included here?

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**enterprise contract (EC)** +
+A set of release policies that you apply to your application snapshot. Use the EC to prevent releases that are not compliant with Supply-chain Levels for Software Artifacts (SLSA) guidelines. 
+
+
```
arewm, line 26: extra return

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A change to one or more files. In {ProductName}, commits that you make to a linked git repository move through the pipeline and automatically get published.
+
+**component** +
+An image that {ProductName} builds from source code in a repository. One or more components that run together form an application.
+
+**Conftest** +
+A utility for testing structured configuration data. Use Conftest to validate container information.
+
+**enterprise contract (EC)** +
+A set of release policies that you apply to your application snapshot. Use the EC to prevent releases that are not compliant with Supply-chain Levels for Software Artifacts (SLSA) guidelines.
```
arewm, line 24: It isn't necessarily limited just to SLSA guidelines, these are just an example of the policies that can be used.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**Conftest** +
+A utility for testing structured configuration data. Use Conftest to validate container information.
+
+**enterprise contract (EC)** +
+A set of release policies that you apply to your application snapshot. Use the EC to prevent releases that are not compliant with Supply-chain Levels for Software Artifacts (SLSA) guidelines. 
+
+
+**integration test** +
+A pipeline that you set up in GitHub. When you add components, it tests each one individually, and then tests the application as a whole.
```
arewm, line 28: @dirgim , is this specific to GitHub?

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**IntegrationTestScenario** +
+A parameter that the integration service uses to test application components.
```
arewm, line 31: A parameter doesn't _feel_ like the right word. An ITS is a kubernetes resource that defines a pipeline to run, I believe.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**Java build service** +
+A group of utilities that accelerates build times, locates dependencies in final images, and meets product security requirements by managing the construction of Java dependencies. 
+
+**lifecycle run** +
+The processes that automatically occur after a merge or pull request is resolved. These processes can include build tests, integration tests, and TaskRuns.
```
arewm, line 37: This confuses me. Where do we use this term? Maybe there is some better way to describe it.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A group of utilities that accelerates build times, locates dependencies in final images, and meets product security requirements by managing the construction of Java dependencies. 
+
+**lifecycle run** +
+The processes that automatically occur after a merge or pull request is resolved. These processes can include build tests, integration tests, and TaskRuns. 
+
+**managed workspace** +
+A workspace that mirrors some details of an already existing workspace. Supporting teams can create a managed workspace to grant limited permissions to development teams in the production environment. 
+
+**OSBS features** +
+The layered image build service, or OSBS (derived from OpenShift Build Service), is a tool for building layered images. With OSBS features, you can build base RHEL images, layered images, and the builder image itself.
```
arewm, line 43: I don't feel like we need this either.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A component that moves all of the relevant PipelineRun information, known as the Pipeline output, to an external storage that is outside of the cluster’s etcd database. The persister runs after the system completes the PipelineRun.
+
+**pipelines as code** +
+A practice that defines pipelines by using source code in Git. Pipelines as Code is also the name of link:https://pipelinesascode.com[a subsystem] that executes those pipelines.
+
+**pipeline results** +
+Systems that retain the history and details of builds. 
+
+**PipelineRun** +
+A collection of TaskRuns that are arranged in a specific order of execution. A valid pipeline is compliant with the customer’s enterprise contract.
```
arewm, line 55: A PipelineRun is a Tekton "thing" and doesn't have anything to do with EC. The PipelineRun will be represented in the SLSA provenance which is then used to evaluate by EC

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**pipeline service** +
+A managed service that securely runs Tekton pipelines. It also stores the metadata and pipeline logs that are related to an executed pipeline in a separate database.
```
arewm, line 58: @gabemontero, is this still relevant for the upstream version of Konflux?

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+
+**product security test** +
+A test that assesses the security of build inputs and outputs with vulnerability scanning, code scanning, image testing, and result reporting. It also identifies outdated images, applications, and build pipelines.
```
arewm, line 61: Calling this a "product" security test seems to bleed out Red Hatisms. Should we call this out specifically as a different type of build-time test? Or just leave them as security-focused build-time tests?

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A managed service that securely runs Tekton pipelines. It also stores the metadata and pipeline logs that are related to an executed pipeline in a separate database.
+
+**product security test** +
+A test that assesses the security of build inputs and outputs with vulnerability scanning, code scanning, image testing, and result reporting. It also identifies outdated images, applications, and build pipelines. 
+
+**pruner** +
+A component that removes resources that are associated with the completed PipelineRuns. The system assigns resources, such as pods, to every PipelineRun. Without a pruner, these resources remain in the cluster indefinitely, even after the system completes the PipelineRun. 
+
+**release pipeline** +
+A pipeline that defines the process for validating snapshots against the enterprise contract. They also provide an array of release destinations.
```
arewm, line 67: It is a generic Tekton pipeline that can be used to perform any configured steps. The EC test that is run is just a pre-requisite that must pass before carrying out the pipeline.

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,98 @@
...
+A component that removes resources that are associated with the completed PipelineRuns. The system assigns resources, such as pods, to every PipelineRun. Without a pruner, these resources remain in the cluster indefinitely, even after the system completes the PipelineRun. 
+
+**release pipeline** +
+A pipeline that defines the process for validating snapshots against the enterprise contract. They also provide an array of release destinations. 
+
+**security testing** +
+A process that determines if images meet security quality standards.
+
+**signature and provenance** +
+Mechanisms that use Tekton chains to extract and store the metadata of the PipelineRuns, sign that metadata, and then store it in the image registry that is next to the component image.
```
arewm, line 73: Is there a reason that these are grouped together? Signatures are cryptographic associations with specific identities. Provenance (i.e. SLSA provenance) is a description of the process that built the artifact.

Konflux uses Chains to generate them, but the signature/provenance don't use chains ... they are an output from Chains.

https://slsa.dev/attestation-model
https://slsa.dev/spec/v1.0/provenance
https://in-toto.readthedocs.io/en/latest/command-line-tools/in-toto-sign.html

arewm (changes requested):

docs/modules/ROOT/pages/glossary/index.adoc
gabemontero: Yeah @arewm @Chr1st1anSears  `pipeline service` as is implied here, i.e. running tekton as a service, is no longer relevant.

I say we lose this section, where we still have the `**Tekton**` section, and then just add and Tekton Results section down below with around were the Tekton and Tekton Chains sections are, that looks like 

```
**Tekton Results** +

A mechanism that stores PipelineRun and TaskRun metadata in a separate database and underlying Pod logs in cloud storage
```

docs/modules/ROOT/pages/glossary/index.adoc
Chr1st1anSears: How about just deleting this definition? I don't think it's necessary for the kind of users that we'll have.

docs/modules/ROOT/pages/glossary/index.adoc
Chr1st1anSears: I'm currently revising the sentence to read, "You can use the EC to prevent..."
I understand that the EC isn't limited to checking SLSA conformance, but surely it's still worth emphasizing the relationship between the two; many EC checks are based on SLSA.

docs/modules/ROOT/pages/glossary/index.adoc
arewm: that works for me.

docs/modules/ROOT/pages/glossary/index.adoc
Chr1st1anSears: I'm gonna remove this definition for now, because I think the new doc about build-time tests does a good job of introducing and defining both build-time tests in general, and highlighting the security tests under that category.

docs/modules/ROOT/pages/glossary/index.adoc
Chr1st1anSears: I'm not sure where that signature and provenance definition came from, but frankly it didn't seem great. I rewrote it, hopefully with better results.

gabemontero (approved):
I'm good with the updates I asked for

docs/modules/ROOT/pages/glossary/index.adoc
```diff
@@ -1 +1,78 @@
...
+A practice that defines pipelines by using source code in Git. Pipelines as Code is also the name of link:https://pipelinesascode.com[a subsystem] that executes those pipelines.
+
+**pipeline results** +
+Systems that retain the history and details of builds. 
+
+**PipelineRun** +
+A collection of TaskRuns that are arranged in a specific order of execution. 
+
+**provenance** +
+{ProductName} produces SLSA provenance, which consists of the attestation for an artifact, and a signature for that attestation. Attestation lists the steps that {ProductName} took to create a given artifact. The signature enables you to verify that no one tampered with that attestation.
```
arewm, line 42: The provenance lists the steps

Chr1st1anSears: Isn't attestation the specific component of provenance that lists the steps? That's how we defined it [here](https://konflux-ci.dev/docs/Konflux-CI/discover/slsa-conformity/#slsa-provenance).

Or would you like me to change this to say:
"Konflux produces SLSA provenance, which includes a list of steps that Konflux took to build a given artifact, and a signature to verify that the provenance comes from Konflux."

arewm: An attestation is generically [a claim being made about a subject](https://github.com/in-toto/attestation?tab=readme-ov-file#in-toto-attestation-framework) (i.e. the container that is produced). The SLSA provenance is a _type_ of attestation but there are many different types of attestations ... some that have been [defined by the community](https://github.com/in-toto/attestation/tree/main/spec/predicates), but others can certainly exist even without being vetted by the in-toto maintainers.

If we are talking about the build steps, that is the provenance attestation .. both here and in https://konflux-ci.dev/docs/#slsa-provenance. I didn't catch that issue in the overview previously.

Chr1st1anSears: Ok, I updated this def and the SLSA overview doc accordingly.

arewm (approved):

PR #52, opened 2024-05-31
nudges: components referenced across applications

lance:
It is possible for components to nudge or be nudged by components in another application altogether. The existing documentation did not make this clear. This change is an attempt to clarify that.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
@@ -1,6 +1,6 @@
 = Defining component relationships
 
-When a workspace includes multiple components, their repositories might contain references to image builds of another component. For instance, an application might include multiple child components that, using a Dockerfile `FROM` clause, reference the image build of a common parent. Or, a component might reference a image builds from another application altogether. OpenShift operators, for example, often have references to several components across applications.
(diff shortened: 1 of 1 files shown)


lance:
My IDE made some whitespace changes as well...

docs/modules/ROOT/pages/how-tos/configuring/component-nudges.adoc
```diff
@@ -11,25 +11,31 @@ However, if you build an application with {ProductName} and make references betw
...
 .Procedure
 
-To define the relationships between components in an application, complete the following steps in the {ProductName} UI:
+To define the relationships between components in a workspace, complete the following steps in the {ProductName} UI:
 
-. Navigate to your application. 
-. On any tab, use the *Actions* drop-down menu to select *Define component relationships*. 
+[NOTE]
+====
+Applications with only one component do not have the option to define relationships in the UI. To achieve the same result for single-component applications, you can define the relationship in the CLI.
```
arewm, line 25: presumably this will be fixed with https://issues.redhat.com/browse/KFLUXBUGS-1313. Should we caveat this to say currently or link the issue? 

@Chr1st1anSears, do you have any thoughts?

Chr1st1anSears: Hmm that's a good question. For external users, I think it'd probably be most helpful just to add "currently," as you suggested.

arewm: @lance, would you be able to make that change?

arewm (approved):

PR #59, opened 2024-06-05
Reset the URL to remove additional slugs

arewm:
konflux-ci/docs#56 changed the URL to add additional slugs which was partially fixed with konflux-ci/docs#57. This furthers the process to revert the url modifications while still enabling the quick high-level overview of the "discover" content.


docs/modules/ROOT/pages/how-tos/metadata/provenance.adoc
```diff
@@ -138,5 +138,5 @@ Example output:
 --
 
 == Additional resources
-* Learn about the SLSA framework and xref:/discover/slsa-conformity.adoc[how {ProductName} meets the requirements of SLSA Build Level 3].
+* Learn about the SLSA framework and xref:/index.adoc[how {ProductName} meets the requirements of SLSA Build Level 3].
```
arewm, line 141: How do we make the xref point to the specific section in the index?

PR #66, opened 2024-06-06
KFLUXBUGS-1339: Document ImageRepository creation

sebsoto:
Required or image push will fail

### DIFF
--- a/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
@@ -68,7 +68,7 @@ spec:
       finalizers:
       - spi.appstudio.redhat.com/remote-secrets
       - application.appstudio.redhat.com/finalizer
-      name: "konflux-multibranch-sample-{{.versionName}}"
+      name: "konflux-multibranch-sample-app-{{.versionName}}"
     spec:
       displayName: "konflux-multibranch-sample-{{.versionName}}"
 
@@ -85,7 +85,7 @@ spec:
       - pac.component.appstudio.openshift.io/finalizer
       name: konflux-multibranch-sample-{{.versionName}}
     spec:
-      application: "konflux-multibranch-sample-{{.versionName}}"
+      application: "konflux-multibranch-sample-app-{{.versionName}}"
       componentName: "konflux-multibranch-sample-{{.versionName}}"
       source:
         git:
@@ -160,6 +160,31 @@ event == "push" && target_branch == "v1.0.0"
 ----
 
 * The `appstudio.openshift.io/application` and `appstudio.openshift.io/component` labels must be adjusted to specify the right Application and Component respectively. Failing to do this will cause builds of the pipeline to be associated with the wrong application or component.
+
+=== Create ImageRepositories for each component
+
+This is done automatically when working with the UI, and must be done manually when creating components through CLI means.
+This must be done for each component to authenticate with the output-registry when pushing built images.
+
+[source,yaml]
+.imagerepository.yaml
+----
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ImageRepository
+metadata:
+  labels:
+    appstudio.redhat.com/application: konflux-multibranch-sample-app-v1-0-0
+    appstudio.redhat.com/component: konflux-multibranch-sample-v1-0-0
+  annotations:
+    image-controller.appstudio.redhat.com/update-component-image: 'true'
+  name: imagerepository-for-konflux-multibranch-sample-app-v1-0-0-konflux-multibranch-sample-v1-0-0
+spec:
+  image:
+    name: konflux-multibranch-sample-app-v1-0-0/konflux-multibranch-sample-v1-0-0
+    visibility: public
+----
+
+`visibility` can be set to either `public` or `private`.
 
 == Known limitations
 


docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
...
+metadata:
+  labels:
+    appstudio.redhat.com/application: konflux-multibranch-sample-v1-0-0
+    appstudio.redhat.com/component: konflux-multibranch-sample-v1-0-0
+  name: imagerepository-for-konflux-multibranch-sample-v1-0-0-konflux-multibranch-sample-v1-0-0
+  namespace: konflux-multibranch-sample-template
+spec:
+  image:
+    name: konflux-multibranch-sample-v1-0-0/konflux-multibranch-sample-v1-0-0
+    visibility: public
```
arewm, line 182: Can we note that this value can be private as well so that private repos can be onboarded with private images?

arewm (commented):
@ifireball , would it be possible for the controller to take care of this? If so, would this documentation only be needed until that functionality is implemented?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
 
 * The `appstudio.openshift.io/application` and `appstudio.openshift.io/component` labels must be adjusted to specify the right Application and Component respectively. Failing to do this will cause builds of the pipeline to be associated with the wrong application or component.
 
+=== Create ImageRepositories for each component
+
+This must be done for each component to authenticate with the output-registry when pushing built images.
```
ifireball, line 166: Is this really needed for all cases? What about the case where all the different versions push into the same image repo? This is what the examples were really showing so far.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
...
+.imagerepository.yaml
+----
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ImageRepository
+metadata:
+  labels:
+    appstudio.redhat.com/application: konflux-multibranch-sample-v1-0-0
+    appstudio.redhat.com/component: konflux-multibranch-sample-v1-0-0
+  name: imagerepository-for-konflux-multibranch-sample-v1-0-0-konflux-multibranch-sample-v1-0-0
+  namespace: konflux-multibranch-sample-template
```
ifireball, line 178: We probably don't need to hard-specify the namespace

ifireball (changes requested):

ifireball:
> @ifireball , would it be possible for the controller to take care of this? If so, would this documentation only be needed until that functionality is implemented?

Well, first I'd like to understand if there is a subset of cases where this wouldn't be needed. I don't recall needing to do this when I was testing the sample that the examples here are based on.

Second, if there is a substantial amount of cases where we would need this, then yeah we'd probably want to allow the template to include these resources too.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
 
 * The `appstudio.openshift.io/application` and `appstudio.openshift.io/component` labels must be adjusted to specify the right Application and Component respectively. Failing to do this will cause builds of the pipeline to be associated with the wrong application or component.
 
+=== Create ImageRepositories for each component
```
mmorhun, line 164: It must be done by UI automatically on Component creation.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
...
+
+This must be done for each component to authenticate with the output-registry when pushing built images.
+
+[source,yaml]
+.imagerepository.yaml
+----
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ImageRepository
+metadata:
```
rcerven, line 173: missing also annotation:
image-controller.appstudio.redhat.com/update-component-image: 'true'

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,27 @@ event == "push" && target_branch == "v1.0.0"
...
+  labels:
+    appstudio.redhat.com/application: konflux-multibranch-sample-v1-0-0
```
rcerven, line 175: application and component should be different in the example, so it is more clear how name/imagename should look like

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: Added

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: Added

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: Added

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
rcerven: it isn't label but 'annotations'

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,30 @@ event == "push" && target_branch == "v1.0.0"
...
+.imagerepository.yaml
+----
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ImageRepository
+metadata:
+  labels:
+    appstudio.redhat.com/application: konflux-multibranch-sample-app-v1-0-0
+    appstudio.redhat.com/component: konflux-multibranch-sample-v1-0-0
+    image-controller.appstudio.redhat.com/update-component-image: 'true'
+  name: imagerepository-for-konflux-multibranch-sample-v1-0-0-konflux-multibranch-sample-v1-0-0
```
rcerven, line 179: should reflect change in application name

even though you can create it like this, it would be better to have in example show that by default it is constructed with app/component name

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +161,30 @@ event == "push" && target_branch == "v1.0.0"
...
+spec:
+  image:
+    name: konflux-multibranch-sample-app-v1-0-0/konflux-multibranch-sample-v1-0-0
```
rcerven, line 182: should reflect change in application name

even though you can create it like this, it would be better to have in example show that by default it is constructed with app/component name

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: fixed

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: you are correct, changed

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
sebsoto: not sure I see another place where this is wrong

rcerven: looks good now:
both imagerepository name
and image name
now have there correct application name

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: @mmorhun can we use the `image.redhat.com/image` annotation on the component instead of defining this resource?  What happens if we define the same annotation value on multiple components? Notably when I was testing this, everything seems to have worked fine while all components had the same `output-image` value configure for the pipeline, and only one of them had the annotation defined on it.

What is the criteria that decides when the push secret is made available to a pipeline? Is this by matching component and application names? Does this work differently when having the `ImageRepository` resource v.s. the `image.redhat.com/image` annotation?

rcerven (approved):

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
mmorhun: Yes, we can have one image repository configured for many components. In general you should proceed with single component as usual, and omit any image repository actions for others except setting Soec.ContainerImage to already existing image repository.
Note, that image repository is bound to the first component and removing it will leave all other components without image repository at all. To prevent that, one should add each component into ownerReference of the ImageRepository CR.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
mmorhun: No, we cannot use `image` annotation because it's like a status: it just shows current state. Moreover, it's deprecated and we are on removing it now. `ImageRepository` is the way to go. It has wider functionality and clearer interface.

mmorhun: > What is the criteria that decides when the push secret is made available to a pipeline?

It should be linked to the pipeline service account

mmorhun (approved):

mmorhun:
@sebsoto @ifireball can we get this merged?

ifireball:
> @sebsoto @ifireball can we get this merged?

I don't think so, since this goes completely against the spirit of what this doc is trying to do. Either we decide to add automation around this in the controller and document how to use it, or we recommend reusing the repository that was created by the UI already.

sebsoto:
If that's the case, I would suggest removing this page from the documentation site while you figure out what to do. The process described in this doc is not functional in the state its in.

ifireball:
> If that's the case, I would suggest removing this page from the documentation site while you figure out what to do. The process described in this doc is not functional in the state its in.

Is it? It does not mention changing the `ImageRepository` on the pipeline files, and it does mention creating the first version in the UI. As far as I can tell it does work. Unless I'm missing something.

sebsoto:
> it does mention creating the first version in the UI

You are referring to this correct?
https://github.com/konflux-ci/docs/blob/00a6d87dac90ba01a7fccca0946b00801cfeadc5/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc?plain=1#L17

> It is recommended, however, for the first time a particular repository is on-boarded, to do so manually via the Konflux UI.

In order for all the required ImageRepositories to be created is necessary to be done via the UI for each application and component. My understanding, which may be false, is that this workflow is meant as an alternative to having to do that.

ifireball:
> 
> In order for all the required ImageRepositories to be created, it is necessary to use the UI to create each application and component. My understanding, which may be false, is that this workflow is meant as an alternative to having to do that.

The part that seems false about your understanding is the first part. See this quote by @mmorhun from a previous thread on this PR:

> Yes, we can have one image repository configured for many components. In general you should proceed with single component as usual, and omit any image repository actions for others except setting Soec.ContainerImage to already existing image repository.
Note, that image repository is bound to the first component and removing it will leave all other components without image repository at all. To prevent that, one should add each component into ownerReference of the ImageRepository CR.

So assuming that:

1. All the components that represent different versions of the same code push into the same image repository (And therefore are configured with identical `output-image` values in the PAC fils apart from the tag). 
2. The first component for a particular piece of code is created in the UI.

Then you will only need one `ImageRepository` resource for all those components and it will be created by the UI.

sebsoto:
Understood, and thanks for the detailed response @ifireball. In that case I will close this PR

PR #73, opened 2024-06-12
Hacdocs 965 olm

Chr1st1anSears:


docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
+= Building an Operator Lifecycle Manager (OLM) in {ProductName}
+
+If you are using an link:https://olm.operatorframework.io/docs/[Operator Lifecycle Manager (OLM)] for your application, you can build the OLM in {ProductName}. The first procedure in this document explains how to use {ProductName} to build an OLM in the most basic sense--building its operator and bundle. The second procedure is optional but recommended. It explains how you can automatically update the image references to the operator and operands that should be present in your bundle.
```
ernesgonzalez33, line 3: I would replace `an` for `the` before the OLM link. Also, you don't build "the OLM", you build an operator that is managed by the OLM

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+
+== Builing the operator and the bundle
```
ernesgonzalez33, line 6: Building* typo

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+[NOTE] 
+====
+This procedure assumes that the source code for your OLM operator and bundle, including the Dockerfile, are in the same git repository, per OLM convention.
```
ernesgonzalez33, line 10: If they are building an operator and a bundle, they will have two Dockerfiles at the very least, so I would change `Dockerfile` to plural here

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+====
+
+.Procedure
+
+. In the {ProductName} UI,  xref:../how-tos/creating.adoc[create a new application] for your OLM in {ProductName}.
```
ernesgonzalez33, line 15: Same as before. OLM is the lifecycle manager, not what you build. I would change it for "OLM operator" as it is below, or simply operator

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+. In your new OLM application, xref:../how-tos/creating.adoc[add a new component] for your OLM operator. Be sure to specify the correct path to the operator's Dockerfile within its git repository.
+. Add another component for your OLM bundle. Enter the same URL that you used for the operator, but enter the path to the bundle's Dockerfile.
+. (Optional) If you are using a File-based Catalog (FBC) for your OLM, you must build the FBC as another component in its own, separate application in {ProductName}.
```
ernesgonzalez33, line 18: I find that comma after "in its own" weird

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+. (Optional) You may want to configure {ProductName} to xref:../how-tos/configuring/redundant-rebuilds.adoc[prevent redundant rebuilds] for this application. For example, you can configure {ProductName} to rebuild your bundle image only if a commit is made to its Dockerfile or the `/bundle` directory, instead of rebuilding it whenever any commit is made to the OLM git repository.
```
ernesgonzalez33, line 19: I would change `the OLM git repository` for `your operator git repository`

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+. Add another component for your OLM bundle. Enter the same URL that you used for the operator, but enter the path to the bundle's Dockerfile.
+. (Optional) If you are using a File-based Catalog (FBC) for your OLM, you must build the FBC as another component in its own, separate application in {ProductName}.
+. (Optional) You may want to configure {ProductName} to xref:../how-tos/configuring/redundant-rebuilds.adoc[prevent redundant rebuilds] for this application. For example, you can configure {ProductName} to rebuild your bundle image only if a commit is made to its Dockerfile or the `/bundle` directory, instead of rebuilding it whenever any commit is made to the OLM git repository. 
+
+== Defining relationships between OLM components
+
+.Procedure
+. In the {ProductName} UI, in your OLM's application, go to the *Components* tab and copy the URL for the operator's container image. The URL should include `sha256:`.
+. Using your preferred text editor, in the git repo for your OLM, open the cluster service version (CSV) file for your bundle. In that file, update the image reference to the operator to be the URL you just copied. Commit this change.
```
ernesgonzalez33, line 25: Same here with OLM.

ernesgonzalez33 (changes requested):

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
Chr1st1anSears: In this case, `an` would be appropriate, since the instructions apply to any Operator you want to build in Konflux. However, you were clearly right about the OLM vs Operator distinction. I applied your other suggested changes. Please let me know if the language looks better now.

(The main remaining challenge was to differentiate between "Operator" meaning the whole package, and "operator" meaning the controller. So I called the single component that gets made into an image the "controller," and referred to the whole application as an "Operator," which I think is consistent with the Operator Framework: https://operatorframework.io/)

ernesgonzalez33: I still don't see it here! Because you are not saying "an operator", but "an Operator Lifecycle Manager". Maybe you could say something like: "If you are using the Operator Lifecycle Manager (OLM) for your operator, you can build it and its bundle in Konflux". Or something along those lines. Because here you are saying that you are using "an" OLM, and there are not several OLMs out there. And after that you say "build the OLM", and that means building the framework itself. At least that's what I understand

ernesgonzalez33: Forget my last comment. I saw the updated paragraph!

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+[NOTE] 
+====
+This procedure assumes that the source code for your Operator's controller and bundle, including the Dockerfiles, are in the same git repository, per OLM convention.
+====
+
+.Procedure
+
+. In the {ProductName} UI,  xref:../how-tos/creating.adoc[create a new application] for your Operator in {ProductName}.
+. In your new Operator application, xref:../how-tos/creating.adoc[add a new component] for your controller. Be sure to specify the correct path to the operator's Dockerfile within its git repository.
```
ernesgonzalez33, line 16: I also don't like the use of "controller" here. An operator image could have several controllers to manage all the operands they need to manage. The difference between a bundle and an operator image is that the first one includes the second one and all the manifests needed to deploy it. Here we are building the operator image, which can have several controllers inside

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,27 @@
...
+. Add another component for your bundle. Enter the same URL that you used for the controller, but enter the path to the bundle's Dockerfile.
```
ernesgonzalez33, line 17: Same here. The "controller" word could make teams believe that they might need to build their controllers one by one and that's not the case

ernesgonzalez33 (changes requested):

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
ernesgonzalez33: @Chr1st1anSears I made some comments. I like the changes but using "controller" as the word for the operator image is not correct either. Because an operator could have several controllers. I commented two of those but there are more in the docs. 

Operator: is the software that deploys and manages the lifecycle of complex applications in Kubernetes clusters. It could include several controllers when there are several operands to manage.

Bundle: image that includes the operator and all the manifests needed to deploy it

docs/modules/ROOT/pages/advanced-how-tos/building-olm.adoc
```diff
@@ -0,0 +1,41 @@
+= Building an Operator Lifecycle Manager (OLM) Operator in {ProductName}
+
+If you are developing an application that aligns with the link:https://operatorframework.io/[Operator Framework], and managing it with link:https://olm.operatorframework.io/docs/[Operator Lifecycle Manager (OLM)], you can build that application in {ProductName}. We refer to such applications as OLM Operators. 
+
+OLM Operators include the following elements:
+
+* Operator, which manages Operands
+* Operands, which are custom resources
```
ernesgonzalez33, line 8: components or resources that the operator manages. Better than just saying "custom resources"

PR #87, opened 2024-07-17
Add doc: Passing build arguments

tkdchen:
STONEBLD-2498


docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,42 @@
...
+       runAfter:
+       - prefetch-dependencies
+       taskRef:
+--
+
+(1) Replace with real build arguments.
+
+(2) Define a Pipeline parameter to accept the passed-in build arguments.
+
+(3) Pass the argument to the link:https://github.com/konflux-ci/build-definitions/tree/main/task/buildah/[buildah task].
```
chmeliik, line 40: It's not clear where in the pipeline YAML these should be done. If I'm not mistaken,
* defining the param (step 2) should be done in `.spec.pipelineSpec.params`
* passing the values (step 1) in `.spec.params`
* and passing the param to the buildah task (step 3) in `.spec.pipelineSpec.tasks[]` where the resolver references the buildah task

It would be helpful to document this and/or to show more context in the code example

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,42 @@
...
++      default: []
+     results:
+     - description: ""
+       name: IMAGE_URL
+@@ -231,6 +236,8 @@
+         value: $(tasks.clone-repository.results.commit)
+       - name: BUILD_ARGS_FILE
+         value: $(params.build-args-file)
++      - name: BUILD_ARGS       (3)
++        value: ["$(params.build-args[*])"]
```
chmeliik, line 30: It might also be good to show how to pass a dynamically generated value (e.g. an array result of a custom task), since that's the more useful use case for BUILD_ARGS

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,42 @@
+= Passing build arguments
```
arewm, line 1: You need to add this page into the closest `_nav.adoc` as well so that it shows up in the left navigation bar. i.e. [how-tos/_nav.adoc](https://github.com/konflux-ci/docs/blob/main/docs/modules/ROOT/pages/how-tos/_nav.adoc?plain=1).

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+Konflux buildah task accepts link:https://github.com/konflux-ci/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L78[`BUILD_ARGS` parameter] that allows passing build arguments to the underlying `buildah` command via `--build-arg` flag. To do this, make the following changes to the PipelineRuns under the `.tekton/` directory:
```
arewm, line 3: Use `{ProductName}` instead of Konflux

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
chmeliik: > to show more context in the code example

Maybe something like the examples in https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#prefetching-pip-dependencies-from-custom-index-servers

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
tkdchen: Updated.

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
tkdchen: Updated.

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
tkdchen: Updated. PTAL whether that's what you meant.

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
tkdchen: Updated.

mkosiarc (approved):

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
chmeliik: I think we could have a more complex example. Maybe add an "Advanced" section at the end, something like

### Advanced

You can do more creative things with the `BUILD_ARGS` parameter. For example, here's a snippet of a pipeline that allows passing build-args via the pipeline parameter (as described above) but also incorporates build-args generated dynamically by other tasks:

```yaml
spec:
  # ...
  pipelineSpec:
    tasks:
    # ...
    - name: build-container
      params:
      # ...
      - name: BUILD_ARGS
        value:
        # pass the build args array from the pipeline param
        - $(params.build-args[*])
        # and a dynamically generated array of build args
        - $(tasks.my-build-arg-generator-task.results.BUILD_ARGS[*])
        # and a build arg with a static name and dynamic value
        - VERSION=$(tasks.my-get-version-task.results.VERSION)
```

docs/modules/ROOT/pages/how-tos/_nav.adoc
```diff
@@ -9,6 +9,7 @@
 *** xref:how-tos/configuring/component-nudges.adoc[Defining component relationships]
 *** xref:how-tos/configuring/rerunning.adoc[Retriggering build pipelines]
 *** xref:how-tos/configuring/redundant-rebuilds.adoc[Preventing redundant rebuilds]
+*** xref:how-tos/configuring/build-with-args.adoc[Passing build arguments]
```
chmeliik, line 12:
```suggestion
*** xref:how-tos/configuring/build-with-args.adoc[Passing buildah build-args]
```

I would call it something like this. "Passing build arguments" is very generic, when I read it the first thing that comes to mind is "oh, this is probably about Tekton params"

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,56 @@
+= Passing build arguments
+
+{ProductName} buildah task accepts link:https://github.com/konflux-ci/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L78[`BUILD_ARGS` parameter] that allows passing build arguments to the underlying `buildah` command via `--build-arg` flag. To do this, make the following changes to the PipelineRuns under the `.tekton/` directory:
+
+. Define a Pipeline parameter in `.spec.pipelineSpec.params` to accept the passed-in build arguments.
```
chmeliik, line 5: I think the default pipeline already has this param.
```suggestion
. Define a `build-args` Pipeline parameter (if not already present) in `.spec.pipelineSpec.params` to accept the passed-in build arguments.
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,56 @@
...
+  # ...
+  pipelineSpec:
+    params:
+    # ...
+    - name: build-args
+      type: array
+      default: []
+--
+
+. Pass the argument to link:https://github.com/konflux-ci/build-definitions/tree/main/task/buildah/[buildah task]. Add `BUILD_ARGS` to `build-container.params`:
```
chmeliik, line 20: same here, the default pipeline does have it

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
...
+      # ...
+      - name: BUILD_ARGS
+        value: ["$(params.build-args[*])"]
+--
+
++
+The value can also be in form `tasks.<taskName>.results.<resultName>[*]` if it is generated by another task dynamically.
+
++
+An advanced example of it would be, you can do more creative things with the `BUILD_ARGS` parameter like this snippet of a pipeline that allows passing `build-args` via the pipeline parameter (as described above) but also incorporates `build-args` generated dynamically by other tasks:
```
chmeliik, line 41: I think the advanced example would work better at the end of the doc. Currently, step 2 is just a too long, and step 3 goes from an advanced example back to something very simple. I'll leave it up to you though

chmeliik (approved):
One more suggestion for re-ordering, but LGTM

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
tkdchen: Thanks for your suggestion. I'd like to keep it here in order to be compact.

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
+= Passing buildah arguments
+
+{ProductName} buildah task accepts link:https://github.com/konflux-ci/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L78[`BUILD_ARGS` parameter] that allows passing build arguments to the underlying `buildah` command via `--build-arg` flag. To do this, make the following changes to the PipelineRuns under the `.tekton/` directory:
+
+. Define a `build-args` Pipeline parameter (if not already present) in `.spec.pipelineSpec.params` to accept the passed-in build arguments.
```
gtrivedi88, line 5:
```suggestion
. Add a `build-args` Pipeline parameter (if not already present) in `.spec.pipelineSpec.params` to accept the passed-in build arguments.
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
...
+        # pass the build args array from the pipeline param
+        - $(params.build-args[*])
+        # and a dynamically generated array of build args
+        - $(tasks.my-build-arg-generator-task.results.BUILD_ARGS[*])
+        # and a build arg with a static name and dynamic value
+        - VERSION=$(tasks.my-get-version-task.results.VERSION)
+--
+
++
+Please note that the value of `BUILD_ARGS` is in the correct syntax described in link:https://tekton.dev/docs/pipelines/variables/[Variable Substitutions Supported by Tasks and Pipelines].
```
gtrivedi88, line 65:
```suggestion
NOTE: Ensure that the value of `BUILD_ARGS` is has the correct syntax as described in link:https://tekton.dev/docs/pipelines/variables/[Variable Substitutions Supported by Tasks and Pipelines].
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
gtrivedi88, line 41:
```suggestion
For example, you can do more creative things with the `BUILD_ARGS` parameter like this snippet of a pipeline that allows passing `build-args` via the pipeline parameter (as described above) but also incorporates `build-args` generated dynamically by other tasks:
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+. Pass desired build arguments in the PipelineRun `.spec.params`:
```
gtrivedi88, line 67:
```suggestion
. Pass the desired build arguments in the PipelineRun `.spec.params`:
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
...
+
++
+[source,yaml]
+--
+spec:
+  params:
+  # ...
+  - name: build-args
+    value: ["ARG1=val", "ARG2=val"]
```
gtrivedi88, line 76:
```suggestion
    # Replace with real argument name and value of each build argument.
    
    value: ["ARG1=val", "ARG2=val"]
```

docs/modules/ROOT/pages/how-tos/configuring/build-with-args.adoc
```diff
@@ -0,0 +1,80 @@
...
+--
+
++
+Replace with real argument name and value of each build argument.
```
gtrivedi88, line 80:
```suggestion

```

gtrivedi88 (approved):

tkdchen:
@gtrivedi88 Thanks for your comments. The doc is updated. PTAL.

gtrivedi88 (approved):

tkdchen:
@arewm PTAL.

PR #101, opened 2024-07-26
Add Troubleshooting docs

chmeliik:
One entry so far: "No space left on device"


chmeliik:
@arewm @TanyaShearonRH PTAL

In the internal Slack, there was some discussion about where to put Troubleshooting docs (split up into related docs vs one unified doc). Let's maybe discuss here instead.

FWIW I'm not a fan of splitting these. If see a failure, I'll go to a troubleshooting doc or just search for it (if the built-in search is good enough), I won't go searching for the doc that is relevant to what I was doing when I saw the failure

arewm (approved):
I am fine with these changes. The point about using the search is fair as well. We have a lot of topics and many troubleshooting items will span across multiple relevant topics.

I am fine with creating a dedicated troubleshooting section for now. We can change it later if needed.

PR #109, opened 2024-08-08
doc(RHTAPWATCH-1105): Update multi-version doc

ifireball:
Update the doc to include all the updates made to the
`project-controller` recently as well as lessons learned from having
users  try it out.

In particular, this update includes instructions for users to create
`ImageRepository`resources for thier components. This therefore,
 consists of a solution to KFLUXBUGS-1339.

### DIFF
--- a/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
@@ -1,24 +1,24 @@
 = Managing multiple software versions
 
-Software projects are often required to maintain multiple versions of the software in parallel in order to support different users of the software with different needs. As far as source code goes, the typical way to maintain multiple versions is by using different Git branches. Using branches in conjunction with {ProductName} can be somewhat tedious as a separate Component needs to be defined for each branch and a separate Application needs to be defined for each collection of components that need to be tested and released together. In addition mutiple other resourced such as *ImageRepositorey*, *IntegrationTestScenario* and *ReleasePlan* need to be defined for each Component or Applicaiton to enable a full CI/CD process.
+Software projects are often required to maintain multiple versions of the software in parallel in order to support different users of the software with different needs. As far as source code goes, the typical way to maintain multiple versions is by using different Git branches. Using branches in conjunction with {ProductName} can be somewhat tedious as a separate Component needs to be defined for each branch and a separate Application needs to be defined for each collection of components that need to be tested and released together. In addition multiple other resources such as *ImageRepository*, *IntegrationTestScenario*, and *ReleasePlan* need to be defined for each Component or Application to enable a full CI/CD process.
 
 The {ProductName} Project Controller seeks to streamline the process of managing multiple versions by introducing the following objects and concepts:
 
 * A *Project* is used to describe a major piece of software that can be worked on by multiple teams over an extended period of time. A project may contain one or more development streams.
 * A *ProjectDevelopmentStream* indicates an independent stream of development. A *ProjectDevelopmentStream* can contain one or more *Applications* each containing one or more *Components*.
-* As described above, starting a new development stream involves creating a large  amount of  *Application*, *Component* and other resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
+* As described above, starting a new development stream involves creating a large amount of *Application*, *Component*, and other resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
 
 == Using the Project Controller
 
-In the sections below we will describe how to use the custom resources supported by the Project Controller to manage multiple software versions and generate Application, Component and other resources.
+In the sections below we will describe how to use the custom resources supported by the Project Controller to manage multiple software versions and generate Application, Component, and other resources.
 
 === Before you begin
 
 The typical process of onboarding a componnent to {ProductName} includes having the {ProductName} Build Service generate a PR to the component Git repository which includes the Tekton Pipeline-as-Code (PaC) pipeline definition for building that component. As a part of this process various other resources such as the PaC Repository resource and Secret resources for accessing the repository are also created.
 
-Whe using the Project Controller to create components the process of generating the PR is skipped because it is assumed the branches being created in the component Git repositores already contain PaC pipeline files. Eert Git repository, however, needs to undergo this process at least once, therefore it is recommended to onboard at least once component from each repository from the UI or by other means before attempting to manage components in that repository via the Project Controller.
+When using the Project Controller to create components the process of generating the PR is skipped because in this document we assume the branches being created in the component Git repositories already contain PaC pipeline files. Each Git repository, however, needs to undergo this process at least once, therefore it is recommended to onboard at least once component from each repository from the UI or by other means before attempting to manage components in that repository via the Project Controller.
 
-The rest of this document is going to assume that process is being followed and would, on occasion requst that data would be copied from the component that was already on-boarded by other means.
+The rest of this document is going to assume that process is being followed and would, on occasion request that data would be copied from the component that was already on-boarded by other means.
 
 === Creating a Project
 
@@ -106,7 +106,7 @@ spec:
 
 The *resources* section for the template may be created by looking at the YAML for existing resources and copying it while removing generated and unnecessary data and adding variable references where needed.
 
-The following kinds of resources are currentyl supported. Please do not attempt to create other kinds as that would prevent the template from being applied successfully.
+The following kinds of resources are currently supported. Please do not attempt to create other kinds as that would prevent the template from being applied successfully.
 
 [cols="2,1,3"]
 |===
@@ -126,18 +126,18 @@ The following kinds of resources are currentyl supported. Please do not attempt 
 
 |appstudio.redhat.com/v1beta2
 |IntegrationTestScenario
-|Note the `v1beta2` API version. Make sure to use the right version when querying the cluster, otherwise important data may be missing from the resutls you get.
+|Note the `v1beta2` API version. Make sure to use the right version when querying the cluster, otherwise important data may be missing from the results you get.
 
 |appstudio.redhat.com/v1alpha1
 |ReleasePlan
 |
 |===
 
-*Note:* When in doubt - the source of truth about what the YAML for your resources should look like is the up to date YAML you can find on the Konflux clusters. We try to keep the exaples here up to date, but with this being a static document there may be a gap between what you see here and what you should configure in tour system. It is recommended that you follow the process of creating the template by copying the YAML of the resrouces you have and cleaning it as describe blow, and not by copying the sample you see here.
+*Note:* When in doubt - the source of truth about what the YAML for your resources should look like is the version you can find on the {ProductName} clusters. We try to keep the examples here up to date, but with this being a static document there may be a gap between what you see here and what you should configure in your system. It is recommended that you follow the process of creating the template by copying the YAML of the resources you have and cleaning it as described below, and not by copying the sample you see here.
 
 Here are specific examples for how to clean up and use the YAML for certain resource kinds:
 
-* For any kind of resource, the `namespace`, `creationTimestamp`, `generation`, `resourceVersion`, `uid`, `ownerReferences` and `finalizers` metadata fields should be removed as well as the `status` section.
+* For any kind of resource specifid in the ProjectDevelopmentStreamTemplate, the `namespace`, `creationTimestamp`, `generation`, `resourceVersion`, `uid`, `ownerReferences`, and `finalizers` metadata fields should be removed as well as the `status` section.
 * For *Application* resources the `metadata.name` and `spec.displayName` fields should contain variable references.
 * For *Component* resources:
 ** The following deprecated annotations should be removed:
@@ -150,7 +150,7 @@ Here are specific examples for how to clean up and use the YAML for certain reso
 ** The `build.appstudio.openshift.io/status` annotation is in place to make the UI present the component's pipeline as customized. Its not mandatory if you can ignore the pipeline status being missrepresented in the UI. Note that this annotation's value needs to contain a reference to a merged PR that added the PaC pipeline files. There is no harm in having multiple components reference the same PR, as long as the `pac.component.appstudio.openshift.io/finalizer` finalizer is not added to the component's finalizers list.
 * For *ImageRepository* resources:
 ** The labels referring to the owning component and application should probably contain variable references.
-** To allow for correct ownership configuration between Component and ImageRepositorey resources, every componet that has its built image pushed to the default registry organisation that is managed automatically by {ProductName}, needs to have a Dedicated ImageRepositorey resource. This implies each component has its own container image repository.
+** To allow for correct ownership configuration between Component and ImageRepository resources, every component that has its built image pushed to the default registry organization that is managed automatically by {ProductName} needs to have a Dedicated ImageRepository resource. This implies each component has its own container image repository.
 ** To meet the above requirement the `spec.image.name` field should contain a variable reference.
 * For *IntegrationTestScenario* resources:
 ** Make sure you query for the `v1beta2` version of those resources, and that you specify that version in your template.
@@ -210,7 +210,7 @@ event == "push" && target_branch == "v1.0.0"
 
 * The value for the `output-image` parameter should be set to match the vale of the `spec.image.name` field of the `ImageRepository` resource that corresponds to the component the pipline would build. 
 +
-If you template is setup in such a way that the repo branch name would appear in the image name, you can use the `target_branch` PaC variable like so:
+If your template is setup in such a way that the repo branch name would appear in the image name, you can use the `target_branch` PaC variable like so:
 +
 [source]
 ----
@@ -218,7 +218,7 @@ If you template is setup in such a way that the repo branch name would appear in
     value: quay.io/redhat-user-workloads/my-tenant/my-app-{{target_branch}}/my-comp-{{target_branch}}:{{revision}}
 ----
 +
-Please note that the examples in this document to not allow for this setup because the image names conain the hyphenized version value while the branch names contain the non-hyphenized value.
+Please note that the examples in this document do not allow for this setup because the image names contain the hyphenated version value while the branch names contain the unhyphenated value.
 
 == Known limitations
 
--- a/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
@@ -14,7 +14,7 @@ In the sections below we will describe how to use the custom resources supported
 
 === Before you begin
 
-The typical process of onboarding a componnent to {ProductName} includes having the {ProductName} Build Service generate a PR to the component Git repository which includes the Tekton Pipeline-as-Code (PaC) pipeline definition for building that component. As a part of this process various other resources such as the PaC Repository resource and Secret resources for accessing the repository are also created.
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+            url: 'https://bombino.preprod.api.redhat.com/v1/sbom/quay/push'
+          event: repo_push
+          method: webhook
+          title: SBOM-event-to-Bombino
+
 ----
 
 The *resources* section for the template may be created by looking at the YAML for existing resources and copying it while removing generated and unnecessary data and adding variable references where needed.
 
+The following kinds of resources are currentyl supported. Please do not attempt to create other kinds as that would prevent the template from being applied successfully.
```
arewm, line 109:
```suggestion
The following kinds of resources are currently supported. Please do not attempt to create other kinds as that would prevent the template from being applied successfully.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -1,25 +1,24 @@
...
+Software projects are often required to maintain multiple versions of the software in parallel in order to support different users of the software with different needs. As far as source code goes, the typical way to maintain multiple versions is by using different Git branches. Using branches in conjunction with {ProductName} can be somewhat tedious as a separate Component needs to be defined for each branch and a separate Application needs to be defined for each collection of components that need to be tested and released together. In addition mutiple other resourced such as *ImageRepositorey*, *IntegrationTestScenario* and *ReleasePlan* need to be defined for each Component or Applicaiton to enable a full CI/CD process.
```
arewm, line 3:
```suggestion
Software projects are often required to maintain multiple versions of the software in parallel in order to support different users of the software with different needs. As far as source code goes, the typical way to maintain multiple versions is by using different Git branches. Using branches in conjunction with {ProductName} can be somewhat tedious as a separate Component needs to be defined for each branch and a separate Application needs to be defined for each collection of components that need to be tested and released together. In addition multiple other resources such as *ImageRepository*, *IntegrationTestScenario*, and *ReleasePlan* need to be defined for each Component or Application to enable a full CI/CD process.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -1,25 +1,24 @@
...
+* As described above, starting a new development stream involves creating a large  amount of  *Application*, *Component* and other resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
 
 == Using the Project Controller
 
-In the sections below we will describe how to use the custom resources supported by the Project Controller to manage multiple software versions and generate Application and Component resources.
+In the sections below we will describe how to use the custom resources supported by the Project Controller to manage multiple software versions and generate Application, Component and other resources.
```
arewm, line 13:
```suggestion
In the sections below we will describe how to use the custom resources supported by the Project Controller to manage multiple software versions and generate Application, Component, and other resources.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -1,25 +1,24 @@
...
 * A *ProjectDevelopmentStream* indicates an independent stream of development. A *ProjectDevelopmentStream* can contain one or more *Applications* each containing one or more *Components*.
-* As described above, starting a new development stream involves creating a large  amount of  *Application* and *Component* resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
+* As described above, starting a new development stream involves creating a large  amount of  *Application*, *Component* and other resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
```
arewm, line 9:
```suggestion
* As described above, starting a new development stream involves creating a large amount of *Application*, *Component*, and other resources. The Project Controller helps to streamline that by allowing to create a *ProjectDevelopmentStreamTemplate* resource that specifies the resources to be created and allows for using variables to customize them. Using a template, many similar development streams can be created quickly.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -1,25 +1,24 @@
...
+The typical process of onboarding a componnent to {ProductName} includes having the {ProductName} Build Service generate a PR to the component Git repository which includes the Tekton Pipeline-as-Code (PaC) pipeline definition for building that component. As a part of this process various other resources such as the PaC Repository resource and Secret resources for accessing the repository are also created.
 
-When creating components the Project Controller assumes that:
+Whe using the Project Controller to create components the process of generating the PR is skipped because it is assumed the branches being created in the component Git repositores already contain PaC pipeline files. Eert Git repository, however, needs to undergo this process at least once, therefore it is recommended to onboard at least once component from each repository from the UI or by other means before attempting to manage components in that repository via the Project Controller.
```
arewm, line 19:
```suggestion
When using the Project Controller to create components the process of generating the PR is skipped because the build service assumes the branches being created in the component Git repositories already contain PaC pipeline files. Each Git repository, however, needs to undergo this process at least once, therefore it is recommended to onboard at least once component from each repository from the UI or by other means before attempting to manage components in that repository via the Project Controller.
```
I think that this wording is confusing though. Would it be better to say something like this?
```suggestion
When new components are generated with the Project Controller, PRs adding the Tekton PipelineRuns needed by PAC are not created. In order to properly trigger pipelines on push and pull request events to the repository, these definitions need to be added my some other means. It is recommended that one branch is onboarded to {ProductName} without the Project Controller to get these files pushed. These definitions can then be copied to other branches that are onboarded to the Project Controller.
```
Will there be a PR to rectify the component names for the branches that have the Tekton files copied from another branch (with another component name)? Or will the PipelineRun be updated with new triggers for the branch, or will users need to perform this manually as well?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -1,25 +1,24 @@
...
 
-* The default pipeline is not being used and that each branch in a component's repo contains the necessary Tekton Pipeline-as-Code files to build it.
-* Any credentials required to access the component repo had already been configured as secrets in the {ProductName} workspace.
+The rest of this document is going to assume that process is being followed and would, on occasion requst that data would be copied from the component that was already on-boarded by other means.
```
arewm, line 21:
```suggestion
The rest of this document is going to assume that process is being followed and would, on occasion request that data would be copied from the component that was already on-boarded by other means.
```

What is "that process"? What is making the request that data is copied?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+|Component
+|
+
+|appstudio.redhat.com/v1alpha1
+|ImageRepository
+|Every component should have an accompanying `ImageRepository` resource
+
+|appstudio.redhat.com/v1beta2
+|IntegrationTestScenario
+|Note the `v1beta2` API version. Make sure to use the right version when querying the cluster, otherwise important data may be missing from the resutls you get.
```
arewm, line 129:
```suggestion
|Note the `v1beta2` API version. Make sure to use the right version when querying the cluster, otherwise important data may be missing from the results you get.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+
+|appstudio.redhat.com/v1alpha1
+|ReleasePlan
+|
+|===
+
+*Note:* When in doubt - the source of truth about what the YAML for your resources should look like is the up to date YAML you can find on the Konflux clusters. We try to keep the exaples here up to date, but with this being a static document there may be a gap between what you see here and what you should configure in tour system. It is recommended that you follow the process of creating the template by copying the YAML of the resrouces you have and cleaning it as describe blow, and not by copying the sample you see here.
```
arewm, line 136:
```suggestion
*Note:* When in doubt - the source of truth about what the YAML for your resources should look like is the version you can find on the {ProductName} clusters. We try to keep the examples here up to date, but with this being a static document there may be a gap between what you see here and what you should configure in your system. It is recommended that you follow the process of creating the template by copying the YAML of the resources you have and cleaning it as describe below, and not by copying the sample you see here.
```

What are you referring to specifically here? Just the resources that are mentioned above? Maybe we can reword this a little bit to say something like this?
```suggestion
*Note*: While we will try to keep the samples above current, they are a snapshot in time. In order to ensure that you are initializing the Project Controller with the latest and up to date versions of the resources, you should follow the process below for preparing the controller's resources.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+
 Here are specific examples for how to clean up and use the YAML for certain resource kinds:
 
-* For any kind of resource, the `namespace`, `creationTimestamp`, `generation`, `resourceVersion`, `uid` and `ownerReferences`, metadata fields should be removed as well as the `status` section.
+* For any kind of resource, the `namespace`, `creationTimestamp`, `generation`, `resourceVersion`, `uid`, `ownerReferences` and `finalizers` metadata fields should be removed as well as the `status` section.
```
arewm, line 140:
```suggestion
* For any kind of resource used by the Project Controller, the `namespace`, `creationTimestamp`, `generation`, `resourceVersion`, `uid`, `ownerReferences`, and `finalizers` metadata fields should be removed as well as the `status` section.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
 * For *Application* resources the `metadata.name` and `spec.displayName` fields should contain variable references.
 * For *Component* resources:
-** The following annotations should be removed:
-*** `build.appstudio.openshift.io/status`
+** The following deprecated annotations should be removed:
```
arewm, line 143: Are these annotations that users will see if they get the resources from the cluster today?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
 *** `image.redhat.com/image`
 ** The `spec.containerImage` field should be removed.
 ** The following fields should probably contain variable references:
 *** `spec.application`
 *** `spec.componentName`
 *** `source.git.revision`
+** The `build.appstudio.openshift.io/status` annotation is in place to make the UI present the component's pipeline as customized. Its not mandatory if you can ignore the pipeline status being missrepresented in the UI. Note that this annotation's value needs to contain a reference to a merged PR that added the PaC pipeline files. There is no harm in having multiple components reference the same PR, as long as the `pac.component.appstudio.openshift.io/finalizer` finalizer is not added to the component's finalizers list.
```
arewm, line 150:
```suggestion
** The `build.appstudio.openshift.io/status` annotation makes the {ProductName} UI present the component's pipeline as customized. While it is not functionally required for component onboarding, omitting the annotation will result in a misrepresentation of this property in the UI. Note that this annotation's value needs to contain a reference to a merged PR that added the PaC pipeline files. There is no harm in having multiple components reference the same PR, as long as the `pac.component.appstudio.openshift.io/finalizer` finalizer is not added to the component's finalizers list.
```

Should the Note portion of this bullet be removed and placed elsewhere?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+* For *ImageRepository* resources:
+** The labels referring to the owning component and application should probably contain variable references.
```
arewm, line 152: Is this a recommendation or a requirement? Should we just state that users should do this instead (as we do two bullets below) of using terminology like "should probably contain?"

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+** To allow for correct ownership configuration between Component and ImageRepositorey resources, every componet that has its built image pushed to the default registry organisation that is managed automatically by {ProductName}, needs to have a Dedicated ImageRepositorey resource. This implies each component has its own container image repository.
```
arewm, line 153:
```suggestion
** To allow for correct ownership configuration between Component and ImageRepository resources, every component that has its built image pushed to the default registry organization that is managed automatically by {ProductName} needs to have a Dedicated ImageRepository resource. This implies each component has its own container image repository.
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -63,53 +62,101 @@ spec:
...
+** To meet the above requirement the `spec.image.name` field should contain a variable reference.
+* For *IntegrationTestScenario* resources:
+** Make sure you query for the `v1beta2` version of those resources, and that you specify that version in your template.
+** The `spec.application` field should probably contain variable references.
+* For *ReleasePlan* resources:
+** The `spec.application` field should probably contain variable references.
```
arewm, line 159: Can we just suggest what users should do here instead of stating that the fields should probably contain variable references?

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +208,18 @@ event == "push" && target_branch == "v1.0.0"
 
 * The `appstudio.openshift.io/application` and `appstudio.openshift.io/component` labels must be adjusted to specify the right Application and Component respectively. Failing to do this will cause builds of the pipeline to be associated with the wrong application or component.
 
+* The value for the `output-image` parameter should be set to match the vale of the `spec.image.name` field of the `ImageRepository` resource that corresponds to the component the pipline would build. 
++
+If you template is setup in such a way that the repo branch name would appear in the image name, you can use the `target_branch` PaC variable like so:
```
arewm, line 213:
```suggestion
If your template is setup in such a way that the repo branch name would appear in the image name, you can use the `target_branch` PaC variable like so:
```

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
```diff
@@ -161,6 +208,18 @@ event == "push" && target_branch == "v1.0.0"
...
++
+[source]
+----
+  - name: output-image
+    value: quay.io/redhat-user-workloads/my-tenant/my-app-{{target_branch}}/my-comp-{{target_branch}}:{{revision}}
+----
++
+Please note that the examples in this document to not allow for this setup because the image names conain the hyphenized version value while the branch names contain the non-hyphenized value.
```
arewm, line 221:
```suggestion
Please note that the examples in this document do not allow for this setup because the image names contain the hyphenated version value while the branch names contain the unhyphenated value.
```

arewm (changes requested):

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: The inmortant part of the "PR" creation process the build service does is IMO the creationg of the PaC Repository resource and accompanying secret. Your 2nd suggestion hoever seems to place the empasys on the PaC files which could be obtained by other means. Also the language seems to imply that users would need to copy PaC files between branches. That would not IMO be the typical case. The typical case would be to onboard the `main` branch manually and then create the `v1.0.0` from it. That way, the `v1.0.0` would already containg PaC files (Albait needing sume adjustments). Given that I decided to go with your first suggestion.

For your 2nd question, there is no automated mechanism to create piplien adjustment PRs for the users, and no such mechanism is planned. We have instructions below about how to manually update the pipeline files as needed.

I suppose that with some component naming conventions (E.g. `comp-name-{{branchName}}`) we could come up with a multi-branch PipelineRun files that would not need adjustments and could also possibly render the current project-controller unneccessary (By, e.g creating the component resource it is building if its missing), but that would be out of scope for the current work.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: The process here refers to the paragraph above - the process of manually onboarding the first component from a repo before using the project-controller for the rest.

The requests to copy data would be done by this document, though I suppose we could come up with a more accurate language, since we do not actually ask the user to copy any data, instead we assume data is copied by creating a branch and then needs to be adjusted.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: The template resource includes copeis of other resources with template variable references added in various places. I know that people tend to copy the example here and then try to modify it for their needs. This can end up trying to create resources with outdated settings. I'm trying here to encourage people to create the template by copying the resources they find in the cluster instead.

Your 2nd suggestion does not seem to me to point at the right direction, accepting the first.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: No, this is not about the project controller using the resources, this is about how to specify them in the template.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: no, but they may see them if they have resources that are older then about a month.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: Putting it elsewhere would place it out of context IMO.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: Well that really depends on wheter the Component and Application are being creatd by the same template or by a template at all. For all we know that user can be using a template to create multiple ImageRepository resources for a single Component they crated manually.

docs/modules/ROOT/pages/advanced-how-tos/managing-multiple-versions.adoc
ifireball: No, becasue we cannot assume that in all cases the users would be creating a ReleasePlan that is referring to an application that is being created by the same template.

arewm (changes requested):
Thanks for the updates. I think this will be ready to merge after the update to line 213: https://github.com/konflux-ci/docs/pull/109/files/5d48f5c0fdd097d06f6eb5d51f87c9cb2bad0523#r1710108993

ifireball (commented):
> Thanks for the updates. I think this will be ready to merge after the update to line 213: https://github.com/konflux-ci/docs/pull/109/files/5d48f5c0fdd097d06f6eb5d51f87c9cb2bad0523#r1710108993 

It looks like it was already applied, I just forgot to close the thread.

arewm (approved):

PR #111, opened 2024-08-10
Hyperlink glossary entries and use a description list

donjuardo:


arewm (approved):
Thanks for this PR, I like the improvement!

Do you know if it would be possible to make the anchor for the entry visible or accessible from the rendered page? For example, rendered markdown pages have the link icon when you hover over a title to make it easy to copy the direct link.

donjuardo:
> Thanks for this PR, I like the improvement!
> 
> Do you know if it would be possible to make the anchor for the entry visible or accessible from the rendered page? For example, rendered markdown pages have the link icon when you hover over a title to make it easy to copy the direct link.

The only viable hack I could find was to use headings for each glossary entry (even though a description list would be more semantic). This way every entry is clickable _and_ linkable with an unique ID and with an icon beside it, which is way better for usability.

The only issue is that the entries are very big with a horizontal rule below them, which is not as aesthetically pleasing as before, and every entry appears in the in-page ToC. I tried using a higher-level heading (`<h5>`) but then Antora starts throwing warnings about incorrect heading levels.

I can push my change with the entries as `<h3>` headings if you wanna take a look.

What do you think, @arewm?

arewm:
Hm. Let's keep it as we have it for now. If we find that we need those links more, then we can adjust it.

PR #112, opened 2024-08-11
Update create-release-plan.adoc

zhanglu1122:
4 should be 1
The first one should be RPA, maybe it was changed but not updated.


zhanglu1122:
https://konflux-ci.dev/docs/advanced-how-tos/releasing/create-release-plan/ here is the page after rendering on the site

arewm (approved):
Thanks for catching this!

PR #113, opened 2024-08-12
Add "Using Trusted Artifacts"

lcarva:
Ref: EC-661

### DIFF
--- a/docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
@@ -1,0 +1,96 @@
+= Using Trusted Artifacts
+
+In order to retain the integrity of the build process, {ProductName} provides two modes of
+operation. The first is by ensuring that every Task in the build Pipeline is
+link:https://enterprisecontract.dev/docs/ec-policies/release_policy.html#trusted_task_package[trusted].
+The second relies on Trusted Artifacts to securely allow sharing of data between Tasks. See
+link:https://konflux-ci.dev/architecture/ADR/0036-trusted-artifacts.html[ADR-36] for more
+information.
+
+Today, when a component is created in {ProductName}, the provided build Pipeline uses Trusted
+Artifacts. However, older components might still be using a version of the build Pipeline which does
+not use Trusted Artifacts. This document explains how to identify if Trusted Artifacts are being
+used and, if not, how to start using them.
+
+== Am I using Trusted Artifacts?
+
+To determine if a certain component is using Trusted Artifacts, inspect its build Pipeline. Look for
+the PipelineRun YAML files under the `.tekton` directory in the git repository of the component. If
+the PipelineRun contains a workspace named `workspace`, it is likely NOT using Trusted Artifacts.
+
+== Migrate to Trusted Artifacts
+
+For components already created in {ProductName}, the simplest way to start using Trusted Artifacts
+is via the link:https://github.com/zregvart/build-migration[build-migration] tool. Run the
+`migration.py` script with the path to the `.tekton` directory as the first parameter. For example:
+`python migration.py ~/src/my-component/.tekton`.
+
+If your build Pipeline has been modified in certain ways, e.g. it includes a custom Task, the
+`build-migration` tool may not work as expected. In this case, perform the required modifications
+manually. These are captured in
+link:https://github.com/konflux-ci/build-definitions/blob/main/pipelines/docker-build-oci-ta/patch.yaml[this]
+patch file. Although the patch file cannot be directly applied to your Pipeline, it describes in
+detail all the required changes.
+
+If your build Pipeline includes custom Tasks that require access to the workspace, e.g. it reads the
+component's source code, the Task must be modified to implement the Trusted Artifacts pattern. At a
+high-level this means accessing data created by other Tasks via OCI artifacts instead of a shared
+workspace backed by a PVC (Persistent Volume Claim).
+
+For example, if your build Pipeline includes a custom Task that accesses the component's source code
+via a workspace:
+
+[source]
+--
+- name: test-kustomize-build
+  workspaces:
+    - name: workspace
+      workspace: workspace
+  runAfter:
+    - clone-repository
+  taskSpec:
+    workspaces:
+      - name: workspace
+    steps:
+      - name: kustomize-build
+        image: registry.access.redhat.com/ubi8/ubi:latest
+        workingDir: $(workspaces.workspace.path)/source
+        script: yum install -y make && make kustomize-build
+--
+
+You can modify the Task to use the Trusted Artifacts pattern:
+
+[source]
+--
+- name: test-kustomize-build
+  params:
+    # New parameter to use the Trusted Artifact from the git-clone Task.
+    - name: SOURCE_ARTIFACT
+      value: $(tasks.clone-repository-oci-ta.results.SOURCE_ARTIFACT)
+  runAfter:
+    - clone-repository-oci-ta
+  taskSpec:
+    params:
+      - description: The Trusted Artifact URI pointing to the artifact with the application source code.
+        name: SOURCE_ARTIFACT
+        type: string
+    stepTemplate:
+      volumeMounts:
+        - mountPath: /var/workdir
+          name: workdir
+    steps:
+      # New step to fetch the Trusted Artifact and make it available to the next steps.
+      - name: use-trusted-artifact
+        image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:8391272c4e5011120e9e7fee2c1f339e9405366110bf239dadcbc21e953ce099
+        args:
+          - use
+          - $(params.SOURCE_ARTIFACT)=/var/workdir/source
+      - name: kustomize-build
+        image: registry.access.redhat.com/ubi8/ubi:latest
+        workingDir: /var/workdir/source
+        script: yum install -y make && make kustomize-build
+    volumes:
+      # New volume to store a copy of the source code accessible only to this Task.
+      - name: workdir
+        emptyDir: {}
+--
--- a/docs/modules/ROOT/pages/advanced-how-tos/_nav.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/_nav.adoc
@@ -13,3 +13,4 @@
 ** xref:advanced-how-tos/building-olm.adoc[Building an OLM operator]
 ** xref:advanced-how-tos/managing-multiple-versions.adoc[Managing multiple software versions]
 ** xref:advanced-how-tos/managing-security-fix.adoc[Managing a security fix]
+** xref:advanced-how-tos/using-trusted-artifacts.adoc[Using Trusted Artifacts]
--- a/docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
@@ -5,7 +5,7 @@ To meet your specific needs, you can customize the way that {ProductName} builds
 * Change the parameters of the build pipeline of a component.
 * Extend the pipeline with your own tasks (or delete existing tasks).
 
-For example, you might decide to limit how long {ProductName} saves the images that it builds for pull requests. You can set a limit by changing the value of the `image-expires-after` parameter. Or you can ensure that images are compliant with regulations for your industry, by adding a compliance check as a task to the build pipeline. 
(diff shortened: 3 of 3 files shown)


docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+link:https://github.com/konflux-ci/build-definitions/blob/main/pipelines/docker-build-oci-ta/patch.yaml[this]
+patch file. Although the patch file cannot be directly applied to your Pipeline, it describes in
+detail all the required changes.
+
+If your build Pipeline includes custom Tasks that require access to the workspace, e.g. it reads the
+component's source code, the Task must be modified to implement the Trusted Artifacts pattern. At a
+high-level this means accessing data created by other Tasks via OCI artifacts instead of a shared
+workspace backed by a PVC (Persistent Volume Claim).
+link:https://github.com/redhat-appstudio/o11y/pull/271/files[This] pull request provides various
+examples, one per commit.
```
simonbaird, line 38: Maybe one more section to answer the question, where can I get help if I need it.

simonbaird (commented):
Very nice.

simonbaird (approved):

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
+= Using Trusted Artifacts
+
+Konflux relies on Trusted Artifacts to securely allow sharing of data between Tasks without
```
arewm, line 3:
```suggestion
{ProductName} relies on Trusted Artifacts to securely allow sharing of data between Tasks without
```

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+compromising the build process. See
+link:https://konflux-ci.dev/architecture/ADR/0036-trusted-artifacts.html[ADR-36] for more
+information.
```
arewm, line 6: This statement makes it sound like the use of trusted artifacts is universal. Should we instead indicate that it is one of the operation modes?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+== Am I using Trusted Artifacts?
+
+To determine if a certain component is using Trusted Artifacts, inspect its build Pipeline. Look for
+the PipelineRun YAML files under the `.tekton` directory in the git repository of the component. If
+the PipelineRun contains a workspace named `workspace`, it is likely NOT using Trusted Artifacts.
+
+== Migrate to Trusted Artifacts
+
+For components already onboarded to Konflux, the simplest way to start using Trusted Artifacts is
+via the link:https://github.com/zregvart/build-migration[build-migration] tool. Run the
```
arewm, line 22: Can we move this tool to somewhere within the konflux-ci GitHub org?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
+= Using Trusted Artifacts
+
+Konflux relies on Trusted Artifacts to securely allow sharing of data between Tasks without
+compromising the build process. See
+link:https://konflux-ci.dev/architecture/ADR/0036-trusted-artifacts.html[ADR-36] for more
+information.
+
+Today, when a component is onboarded to Konflux, the provided build Pipeline uses Trusted Artifacts.
```
arewm, line 8: The word "create" is used instead of "onboard" in https://konflux-ci.dev/docs/how-tos/creating/

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
arewm, line 38: Is this just a generic example for switching to OCI artifacts or are you trying to show something specifically in the PR? Is there a suitable example that we can use that is within the konflux-ci namespace?

arewm: We don't really have a way to interact with potential readers other than for them to create issues.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+`build-migration` tool may not work as expected. In this case, perform the required modifications
+manually. These are captured in
+link:https://github.com/konflux-ci/build-definitions/blob/main/pipelines/docker-build-oci-ta/patch.yaml[this]
+patch file. Although the patch file cannot be directly applied to your Pipeline, it describes in
+detail all the required changes.
+
+If your build Pipeline includes custom Tasks that require access to the workspace, e.g. it reads the
+component's source code, the Task must be modified to implement the Trusted Artifacts pattern. At a
+high-level this means accessing data created by other Tasks via OCI artifacts instead of a shared
+workspace backed by a PVC (Persistent Volume Claim).
```
arewm, line 36: Should we state that this can be done by adding parameters and the TA stepaction?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+To determine if a certain component is using Trusted Artifacts, inspect its build Pipeline. Look for
+the PipelineRun YAML files under the `.tekton` directory in the git repository of the component. If
+the PipelineRun contains a workspace named `workspace`, it is likely NOT using Trusted Artifacts.
+
+== Migrate to Trusted Artifacts
+
+For components already onboarded to Konflux, the simplest way to start using Trusted Artifacts is
+via the link:https://github.com/zregvart/build-migration[build-migration] tool. Run the
+`migration.py` script with the path to the `.tekton` directory as the first parameter. For example:
+`python migration.py ~/src/my-component/.tekton`.
```
arewm, line 24: Are there any requirements for this script (i.e. a minimum python version)? I guess that these requirements would be better on the repository's README. Can we also modify that README to show an example of creating the virtual environment, installing dependencies, and running the tool?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
```diff
@@ -0,0 +1,38 @@
...
+Today, when a component is onboarded to Konflux, the provided build Pipeline uses Trusted Artifacts.
+However, older components might still be using a version of the build Pipeline which does not use
+Trusted Artifacts. This document explains how to identify if Trusted Artifacts are being used and,
+if not, how to start using them.
+
+== Am I using Trusted Artifacts?
+
+To determine if a certain component is using Trusted Artifacts, inspect its build Pipeline. Look for
+the PipelineRun YAML files under the `.tekton` directory in the git repository of the component. If
+the PipelineRun contains a workspace named `workspace`, it is likely NOT using Trusted Artifacts.
```
arewm, line 17: Would it be better to check for common workspaces reused across multiple tasks or for the presence/absence of the TA parameters used by the stepactions?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
arewm, line 8: We don't technically have the ability to create components with the UI with this project since the project does not have a way to automatically create the required image repositories.  In order to be consistent with this documentation, we would have to provide the way for users to specify the pipeline in their gitops repo. Can we just state the name of the pipeline to select/configure?

If a pipeline isn't specified by default in the gitops flow, will the build service automatically select the OCI-TA pipeline by default or will it fall back to the older pipeline?

arewm (changes requested):

arewm:
Can we add a section in https://konflux-ci.dev/docs/how-tos/configuring/customizing-the-build/#extending-the-build-pipeline-with-your-own-tasks for extending tasks leveraging OCI-TA as well?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: I was going to write something along the lines of "if you need help with the `build-migration` tool, file an issue on that repo". But given that that's usually how you'd get help from such tools, maybe it's better to leave it out. That would also raise the question of getting help on Konflux in general which seems a bit more ambitious than what this page is trying to achieve. Maybe a top-level page addressing that will be a nice touch in the future.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: Is it not the intended universal operation mode?

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: https://issues.redhat.com/browse/EC-792

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: The PR is converting custom Tasks from using a shared workspace to Trusted Artifacts. Any suggestions on how I can make that clearer?

I'd rather use those examples as they provide different variations which would be helpful to others.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: Agree that that information should reside in that repo. PRs welcome.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: This covers most cases. The cases that do not are the ones where the user has made explicit changes to the Pipeline. I consider those advanced users that should be able to understand this. Adding too many details here can potentially confuse the non-advanced users which is the majority of them.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: Hmm. I realize that I don't understand how onboarding works when the UI is not involved. Is there a "send PR" workflow outside the UI?

lcarva:
> Can we add a section in https://konflux-ci.dev/docs/how-tos/configuring/customizing-the-build/#extending-the-build-pipeline-with-your-own-tasks for extending tasks leveraging OCI-TA as well?

The section after that needs some attention. Even without Trusted Artifacts, we shouldn't be guiding users toward just excluding `trusted_task.trusted` without warning them of the repercussions.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: I removed the examples and added one directly to the doc :shrug:

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
lcarva: Hopefully the newly added (annotated) example makes this clear.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
simonbaird: Sounds good to me. Actually I was thinking about telling RH users to ask for help on slack, but these are upstream docs, so that's not right.

docs/modules/ROOT/pages/advanced-how-tos/using-trusted-artifacts.adoc
arewm: While PVCs are more limited in terms of their customization, they will likely be more efficient and faster if builds involve many GBs of data being transmitted between tasks.

There will be a heavy preference for it to be the default but that doesn't _necessarily_ mean that it will be the only option.

lcarva: Fair point. Rewrote this paragraph to mention the two modes.

arewm (approved):
Approving these changes. The content is an improvement and there are no glaring issues. We can continue iterating and refining the content.

Thanks for the updates!

PR #120, opened 2024-08-26
Document guidelines for repos using git submodules

CryptoRodeo:
JIRA: CWFCONF-10293


CryptoRodeo:
@arewm , @ralphbean FYI

arewm (commented):
Thanks for this. I'll take a closer look later, but I did notice that you haven't included a link to this page on the navigation pane.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
+= Git Submodule pattern
```
arewm, line 1: You are introducing the term "pattern" ... can you explain what you mean by that? Or give background for what this pattern solves?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+== How this workflow pattern works:
+
+1. Create a "downstream" repo.
+2. Track your upstream git repo with a git submodule reference.
+3. Onboard to {ProductName}, use link:https://github.com/renovatebot/renovate[Renovate] to keep your Git submodule references up to date.
+4. In your downstream repo, apply patches in your Containerfile if you need to carry downstream patches.
```
arewm, line 8: Do we need this section? We will get a TOC generated for the right of the pages as an overview.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+== Implementing this solution pattern
+=== Create a downstream repo
+
+A **downstream repo** is a self-owned copy of the upstream repository whose build, test, release lifecycle you want to control completely.
```
arewm, line 13: Maybe it would be better to define upstream and downstream at the start instead of defining downstream in terms of upstream and then later defining upstream.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+The downstream repo will:
+
+* be used to hold the git submodule reference to the upstream repository.
+* contain a Containerfile that references and pulls in the git submodule(s) content.
```
arewm, line 18: This isn't _strictly_ a requirement.

> and pulls in the git submodule(s) content

What do you mean by this?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+* become a Component of our {ProductName} application.
```
arewm, line 19: Maybe just state something like it will be onboarded to {ProductName}? There isn't _necessarily_ only a single Application. The fact that the component belongs to an application is likely irrelevant here.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+----
+
+Check and configure the git submodule config:
+[source, gitmodules]
+----
+# .gitmodules
+[submodule "<project-name>"]
+  path = <path-in-downstream-repo>
+  url = https://github.com/<namespace>/<project>.git
+  branch = main # Ensure the branch is correct
```
arewm, line 39: Is there a branch added by default? If not, should we direct users to specifically add this?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+----
+See the link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on configuring git submodules.
+
+=== Onboard the project onto {ProductName}
+
+After creating the repository, onboard to {ProductName} and configure your application and its components.
```
arewm, line 45: It may not be a new application. I think we can probably just say to onboard the component and xref to the relevant docs.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+**Note:**
+The xref:/getting-started/index.adoc[Getting started docs] are currently geared towards configuring YAML files.
```
arewm, line 48: What do you mean YAML files? i.e. there is no documentation for onboarding via the UI? Instead, the CRs need to be created manually?

I think that that can be left out of _this_ context as it is relevant at the xref target.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+There does not seem to be any detailed steps on how to do this yet via the UI.
+
+Here are some **brief** steps on onboarding to {ProductName} using the UI:
+
+1. Create an application in your workspace.
+2. Configure your application component.
+3. If you're using Github, ensure that you have the link:https://github.com/apps/red-hat-konflux[Red Hat {ProductName} bot] enabled.
+4. Check your {ProductName} CI results. Ensure that your component was able to be pulled from, built and tested.
+
+For more information on configuring creating applications and components, please read xref:/how-tos/creating.adoc[this documentation].
```
arewm, line 63: It is not currently possible to onboard with the UI due to the fact that we do not have a means to create the image repositories. This is a known gap with the upstream project and our own implementation. While it is unfortunate that we are not documenting this flow, I think we can leave it out for now.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+2. **Automatic Dependency Syncing with link:https://github.com/renovatebot/renovate[Renovate]**: By automating the dependency sync process, whether its syncing the git submodule, the Containerfile, etc. your downstream repo can always use the latest code from the upstream repo(s), ensuring that your project benefits from new features, bug fixes, and security updates.
+3. **Version Control for upstream repos**: You can lock the submodule to a specific commit or version, giving you control over when and how updates are integrated. This is especially useful if an upstream update introduces breaking changes.
+4. **{ProductName} CI**: Automated building and testing of your software artifacts help ensure that changes in the upstream repo don't break your downstream project.
+
+== Potential drawbacks
+
+1. **Complexity**: Managing submodules and automating their updates adds complexity to your workflow, such as:
+  - Debugging issues related to submodule updates
+  - Your automation configuration can be more challenging.
+2. **Dependency Management**: Breaking changes in the upstream repo can affect your downstream repo, requiring manual intervention to fix integration issues.
```
arewm, line 77: One of the biggest challenges here is if the build process changes upstream and we have our own copy of the build configuration (i.e. Containerfile) in our repo. There are some specific practices that we can propose to mitigate that. Can we be more prescriptive here ... or at least more cautionary?

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,90 @@
...
+2. Track your upstream git repo with a git submodule reference.
+3. Onboard to {ProductName}, use link:https://github.com/renovatebot/renovate[Renovate] to keep your Git submodule references up to date.
+4. In your downstream repo, apply patches in your Containerfile if you need to carry downstream patches.
+
+== Implementing this solution pattern
+=== Create a downstream repo
+
+A **downstream repo** is a self-owned copy of the upstream repository whose build, test, release lifecycle you want to control completely.
+
+The downstream repo will:
```
arewm, line 15: I forgot to mention that it might be helpful to illustrate to readers how to configure the components. For example, what should their Context directory be? Their Dockerfile path? How would the Containerfile used differ from upstream?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I'm not sure I want to use "pattern" anymore. Do you have any recommendations? Here are some ideas replacements I can think of:
- Guideline
- Approach
- Protocol
- Blueprint

For now I swapped it to "guideline".

I remember in our discussion that we want to emphasize that this is **not** officially supported by Konflux, but I'm still not sure what word would be good...

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I added it as a brief summary to what this guideline achieves, I removed it

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I moved it up to a "Basic definitions" section under "Implementing this guideline"

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Replaced "will" with "can" to emphasize it's not a requirement.

For "and pulls in the git submodule(s) content" I was referencing the [git-clone build definition](https://github.com/konflux-ci/build-definitions/blob/main/task/git-clone/0.1/git-clone.yaml#L223-L231) that @ralphbean mentioned when answering my question (this is in the slack thread I added you to):
> In the Git submodule example, I see the Containerfile referencing the oras git submodule directory:
ENV ORASPKG /oras
>
>Is there some external service that pulls in that git submodule when the Containerfile is being built?
>
>For example, if I pull in this repo locally: https://github.com/konflux-ci/oras-container
and I go to the oras directory, it's empty until I run the following:
git submodule update --init --recursive
>
>I'm wondering at what point does this repo get pulled in so the Containerfile can use it 

The "Git submodule example" being this one: https://github.com/konflux-ci/oras-container

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: changed to "onboard onto {ProductName}"

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I added a section under this about configuring the branch value. It's not specified by default from my experience.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Ack, I referenced the "Creating applications and components" doc

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I removed this part. Most of my experience with Konflux is using the UI. I have yet to do anything via the CLI or define CRs in Konflux.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: removed from doc

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I leaned more cautionary after editing this section, since I do not know what practices to prescribe yet. If you have any recommendations I will add them

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I added some basic examples for a [component definition](https://github.com/konflux-ci/docs/pull/120/files#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R29-R47) and [Dockerfile](https://github.com/konflux-ci/docs/pull/120/files#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R57-R70) using the git submodule contents and making basic edits. I'm not sure if this is what you had in mind

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,134 @@
...
+  - Debugging issues related to submodule updates
+  - Your automation configuration can be more challenging.
+2. **Dependency Management**: Breaking changes in the upstream repo can affect your downstream repo. In scenarios like this it is strongly recommended to proceed with caution when fixing such issues.
+3. **Merge Conflicts**: If you have automatic updates configured, this might lead to merge conflicts, especially if both the upstream and downstream repos are being actively developed. Resolving these conflicts might require manual effort.
+
+== Gitlab support
+
+GitLab is supported, you need to follow xref:/how-tos/configuring/creating-secrets.adoc [Creating secrets for your build] to provide the token to access your GitLab repository.
+
+To do that you have to create a secret from command line. Refer to link:https://gitlab.cee.redhat.com/konflux/docs/users/-/blob/main/topics/getting-started/getting-access.md#accessing-konflux-via-cli[Accessing {ProductName} via CLI] for how to login with CLI.
```
ralphbean, line 128: Unfortunately, we can't link to gitlab.cee docs here, since upstream readers won't be able to follow them.

Once we have https://issues.redhat.com/browse/KONFLUX-3550 done, there will hopefully be a hook for you to *extend* this document downstream with further information like this.

For now, I think it's best to remove the bit about which cluster to use, internal network, etc.

You could move all of that to the downstream docs repo now in a *very* short doc that first just links to this doc and then contains additional downstream details.

CryptoRodeo: I removed the bits about you mentioned, please see: https://github.com/konflux-ci/docs/pull/120/commits/2f9c8128313c7975c8fe6fcdb782647c89c3a363

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+=== Using private repositories as a git submodule
+
+==== Github
+
+When using private repositories we encourage that you give the link:https://github.com/apps/red-hat-konflux[Red Hat {ProductName} Github app] access to that private repository.
+
+==== Gitlab
+
+For Gitlab repositories please see xref:/how-tos/configuring/creating-secrets.adoc#creating-source-control-secrets[Creating secrets for your builds]
```
arewm, line 107: The context for this is that users will need to configure the secret and then these private repositories should "just work"? Can we add that context to this section?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+See the link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on configuring git submodules.
+
+**Note**: By default the branch is not specified in the git submodule configuration file when adding a git submodule.
+see link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on how to configure the branches for the submodules.
+
+=== Onboard the component onto {ProductName}
+
+After creating the downstream repository you can onboard the component. See xref:/how-tos/creating.adoc[Creating applications and components] for more details.
+
+=== Using private repositories as a git submodule
```
arewm, line 99: Since the content for this section is so small, I don't think that we need to define the specific subheadings.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+See xref:/how-tos/creating.adoc[Creating applications and components] for more details.
+
+==== Configuring the Containerfile
+
+In the Containerfile you can specify the git submodule directories and add any necessary edits that differ from the upstream repository. Here is a basic example:
+
+[source, Dockerfile]
+----
+FROM <base-image>
+ENV <SUBMODULEPKG> /path/to/submodule
```
arewm, line 58: Did you mean to `COPY` after this? Or you might have thought that that step doesn't need documentation.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+# Install necessary dependencies
+RUN dnf -y install <dependencies>
+
+# Set the working directory
+WORKDIR ${<SUBMODULEPKG>}
+
+RUN sed -i 's/<old-text>/<new-text>/g' /file/in/submodule/directory
```
arewm, line 66: You only show changing text here. Would it be better to show content that is more generic like

```dockerfile
# Commands to update the submodule for the current image build
RUN sed -i 's/<old-text>/<new-text>/g' file/in/submodule/directory
# Commands to build binary from submodule sources
RUN go build file/in/submodule/directory
```

(Also, removing the leading `/` would mean that you are working within the submodule working directory instead of needing to pass the absolute path).

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+3. **Version Control for upstream repos**: You can lock the submodule to a specific commit or version, giving you control over when and how updates are integrated. This is especially useful if an upstream update introduces breaking changes.
+4. **{ProductName} CI**: Automated building and testing of your software artifacts help ensure that changes in the upstream repo don't break your downstream project.
+
+== Potential drawbacks
+
+1. **Complexity**: Managing submodules and automating their updates adds complexity to your workflow, such as:
+  - Debugging issues related to submodule updates
+  - Your automation configuration can be more challenging.
+2. **Dependency Management**: Breaking changes in the upstream repo can affect your downstream repo. In scenarios like this it is strongly recommended to proceed with caution when fixing such issues.
+3. **Merge Conflicts**: If you have automatic updates configured, this might lead to merge conflicts, especially if both the upstream and downstream repos are being actively developed. Resolving these conflicts might require manual effort.
```
arewm, line 122: How would we get merge conflicts? Consuming changes from upstream would only be contained within the submodule. These shouldn't be affected by any changes made in the downstream repository.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+2. **Automatic Dependency Syncing with link:https://github.com/renovatebot/renovate[Renovate]**: By automating the dependency sync process, whether its syncing the git submodule, the Containerfile, etc. your downstream repo can always use the latest code from the upstream repo(s), ensuring that your project benefits from new features, bug fixes, and security updates.
+3. **Version Control for upstream repos**: You can lock the submodule to a specific commit or version, giving you control over when and how updates are integrated. This is especially useful if an upstream update introduces breaking changes.
+4. **{ProductName} CI**: Automated building and testing of your software artifacts help ensure that changes in the upstream repo don't break your downstream project.
+
+== Potential drawbacks
+
+1. **Complexity**: Managing submodules and automating their updates adds complexity to your workflow, such as:
+  - Debugging issues related to submodule updates
+  - Your automation configuration can be more challenging.
```
arewm, line 120: Do you want to call out the potential for drift between the upstream and downstream repositories?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+The downstream repo can be used to do the following:
+
+* hold the git submodule reference to the upstream repository.
+* use the contents of the git submodule in a Containerfile for further edits.
+* onboarded onto {ProductName}.
+
+==== Configuring components
+
+Here is an example of a component definition:
+
```
arewm, line 27: I think that this is more about defining components to use the proper contexts than it is just about the component definition. Since that is the case, I think that we can also change the content here:

* Describe the process to ensure the context is defined by the UI
* Describe the process to ensure that the context is defined by the CR definition

That being said, I think that this information would be better placed _in_ the documentation for defining the Component and this documentation would just be something like 

```
+
CAUTION: Make sure to properly set the xref:/path/to/doc.adoc#section[context and containerfile path] as required for the artifact build.
```

The context _may_ be just the submodule or it _may_ be `.` depending on whether additional content is defined in the repo (i.e. to help with migration of data).

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I removed this section. This was meant more for general Gitlab support, not the "using private repos" scenario. I have not found a way to do this outside of the Github Konflux application. Maybe we can add a new section once we have a similar app on Gitlab?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Sounds good, this section has been simplified: https://github.com/konflux-ci/docs/pull/120/commits/f1be568ae08b9605e685c3d1e1e288862518855a#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R83

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+    image.redhat.com/generate: '{"visibility": "public"}'
+spec:
+  application: <application-name>
+  componentName: <component-name>
+  source:
+    git:
+      revision: main
+      url: <downstream repo name>
+      dockerfileUrl: <Dockerfile/Containefile path in downstream repository>
+      context: <directory path used to build the Containerfile/Dockerfile>
```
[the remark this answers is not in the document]
CryptoRodeo: I removed this section and referenced the "Creating applications and components" doc based on your other comment. That info should probably go there instead

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I meant to move `ENV` down further, see https://github.com/konflux-ci/docs/pull/120/commits/f1be568ae08b9605e685c3d1e1e288862518855a#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R32-R54

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Added, thanks! -> https://github.com/konflux-ci/docs/pull/120/commits/f1be568ae08b9605e685c3d1e1e288862518855a#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R43-R46

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Potential for drift has been mentioned: https://github.com/konflux-ci/docs/pull/120/commits/f1be568ae08b9605e685c3d1e1e288862518855a#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R99

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+3. **Version Control for upstream repos**: You can lock the submodule to a specific commit or version, giving you control over when and how updates are integrated. This is especially useful if an upstream update introduces breaking changes.
+4. **{ProductName} CI**: Automated building and testing of your software artifacts help ensure that changes in the upstream repo don't break your downstream project.
+
+== Potential drawbacks
+
+1. **Complexity**: Managing submodules and automating their updates adds complexity to your workflow, such as:
+  - Debugging issues related to submodule updates
+  - Your automation configuration can be more challenging.
+2. **Dependency Management**: Breaking changes in the upstream repo can affect your downstream repo. In scenarios like this it is strongly recommended to proceed with caution when fixing such issues.
```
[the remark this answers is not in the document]
CryptoRodeo: One scenario I thought of was files in the upstream repo being moved, messing up any references to them downstream or in the Containerfile. I removed this section for now.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+3. **Merge Conflicts**: If you have automatic updates configured, this might lead to merge conflicts, especially if both the upstream and downstream repos are being actively developed. Resolving these conflicts might require manual effort.
+
+== Gitlab support
+
+GitLab is supported, you need to follow xref:/how-tos/configuring/creating-secrets.adoc [Creating secrets for your build] to provide the token to access your GitLab repository.
+
+After onboarding your downstream project repo in Gitlab onto {ProductName} you should be able to follow a similar guideline to the one outlined here.
```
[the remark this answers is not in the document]
CryptoRodeo: I removed the Gitlab sections from this doc for now since both should be supported and documented in the docs, so there's probably no need to call out Gitlab specifically.

docs/modules/ROOT/pages/how-tos/worklows/index.adoc
```diff
@@ -0,0 +1,6 @@
+= Generic guidelines for {ProductName} integrated repositories
+
+Depending on your use case, there are several guidelines you can use for repositories that are integrated with {ProductName}:
+
+* xref:/how-tos/workflows/git-submodules.adoc[Git Submodule pattern]
```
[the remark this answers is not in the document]
CryptoRodeo: Yeah! Maybe something like:
```adoc
* xref:/how-tos/workflows/git-submodules.adoc[Git Submodule pattern]
** xref:<link>[Preventing drift in Git submodules ]
```
?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Good idea, I removed this section for now and referenced the docs on creating applications and components. Those details should be put there instead.

arewm: Would you mind adding that information to the creating pages and then linking to it?

CryptoRodeo: I do not mind, I'll add it now and link back to it here

docs/modules/ROOT/pages/how-tos/creating.adoc
```diff
@@ -116,6 +116,20 @@ spec:
   url: "https://github.com/konflux-ci/anothertestrepo"
 --
 
+=== Context and Containerfile path
```
CryptoRodeo, line 119: @arewm Here is the added section in creating.adoc. Please let me know if I should add any edits, I wasn't 100% sure what should go here.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I could add a link to your doc as well on this line if you think that's a good idea?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,128 @@
...
+RUN dnf -y install <dependencies>
+
+# Set the working directory
+WORKDIR ${<SUBMODULEPKG>}
+
+RUN sed -i 's/<old-text>/<new-text>/g' /file/in/submodule/directory
+
+WORKDIR /app
+
+ENTRYPOINT ["<entrypoint-executable>"]
```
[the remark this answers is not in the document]
CryptoRodeo: Labels section added: https://github.com/konflux-ci/docs/pull/120/files#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R52-R54

I'm not sure if you would like a comment above it like `# Adding labels in your Containerfile is recommended`

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Good point, this section has been removed.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Doc section linked: https://github.com/konflux-ci/docs/pull/120/files#diff-46f8b5b0ae481a31c84644c1fde822df6be41954ac5d4002de9444fe1b91fd08R26

docs/modules/ROOT/pages/how-tos/creating.adoc
arewm: I think that I will end up addressing this in my PR, i.e. https://github.com/konflux-ci/docs/pull/126/files#diff-bb9886452ec23549827a5ae772b53b3fbc703dd8e7aa5d196516016cfe8bde46R93-R94

Instead of adding it here, would you be able to comment on my PR to remind me to add a reference where needed?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+The downstream repo can be used to do the following:
+
+* hold the git submodule reference to the upstream repository.
+* use the contents of the git submodule in a Containerfile for further edits.
+* onboarded onto {ProductName}.
+
+==== Configuring components
+
+CAUTION: Make sure to properly set the xref:/how-tos/creating.doc#context-and-containerfile-path[Context and Containerfile path] as required for the artifact builds.
```
arewm, line 26: Since I might do this instead, just add the caution without the xref. You can, however, suggest what context might be needed (i.e. keep it at the repo root so that you can have the Contaienrfile and the submodule within the context).

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
+= Git Submodule guideline
```
arewm, line 1: The workflow isn't about git submodules as much as it is about building upstream projects _with_ git submodules. Users may have git submodules in their repositories for other reasons ... would you expect these guidelines to still be relevant?

Your first sentence for motivating this page is
> This document outlines some basic guidelines for synchronizing changes between an upstream and downstream repository using Git submodules.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+This document outlines some basic guidelines for synchronizing changes between an upstream and downstream repository using Git submodules.
+
+By leveraging Git submodules, it is possible to maintain a consistent and up-to-date relationship between related repositories (upstream and downstream) while having more control of how the project is built, tested and released.
+
+Additionally, the integration with Konflux can automate key aspects of this process, ensuring that changes from the upstream repository are efficiently and reliably propagated to the downstream repository.
```
arewm, line 7: What is "the integration"?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+- **Upstream repo**: a community project which you do not have push access to (or which you do not want to push CI changes into).
+
+- **Downstream repo**: a self-owned copy of the upstream repository whose build, test, release lifecycle you want to control completely.
+
+=== Create a downstream repo
+
+The downstream repo can be used to do the following:
+
+* hold the git submodule reference to the upstream repository.
```
arewm, line 20: We should be consistent between repo and repository.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
+# .gitmodules
+[submodule "<project-name>"]
+  path = <path-in-downstream-repo>
+  url = https://github.com/<namespace>/<project>.git
+----
+See the link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on configuring git submodules.
+
+**Note**: By default the branch is not specified in the git submodule configuration file when adding a git submodule.
+see link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on how to configure the branches for the submodules.
```
arewm, line 79: I don't think we need the same link back to back unless you are pointing to a specific location in the submodule documentation. Can we combine these sentences?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+[source, gitmodules]
+----
+# .gitmodules
+[submodule "<project-name>"]
+  path = <path-in-downstream-repo>
+  url = https://github.com/<namespace>/<project>.git
+----
+See the link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on configuring git submodules.
+
+**Note**: By default the branch is not specified in the git submodule configuration file when adding a git submodule.
```
arewm, line 78: Should we suggest that they set the branch (or inform them that they can set the branch)? If we inform them that they should set the branch, we can show it as an example and then call out that ti isn't added by default.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
```diff
@@ -0,0 +1,101 @@
...
+see link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on how to configure the branches for the submodules.
+
+=== Onboard the component onto {ProductName}
+
+After creating the downstream repository you can onboard the component. See xref:/how-tos/creating.adoc[Creating applications and components] for more details.
+
+=== Using private repositories as a git submodule:
+
+When using private repositories we encourage that you give the link:https://github.com/apps/red-hat-konflux[Red Hat {ProductName} Github app] access to that private repository.
```
arewm, line 87: This is only scoped for GitHub. GitLab private submodules are supported as well ... as long as the PAT has access to the repo?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Let's leave the link out for now. We haven't linked to that document yet from the konflux-ci.dev/docs site ... unless you feel like we should. We can also add the information to the docs here and link to it from my work.

arewm (changes requested):

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Sounds good, the changes have been applied here: https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R26-R27

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I have updated the title, see https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R1

Does that seem better?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I removed "the integration" and left it as "Konflux" so it's less confusing

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Good catch, I updated all "repo" references to the full name now

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Additional link has been removed, please see: https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R83

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I added a section demonstrating how the branches are added: https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R76-R80

Does that seem good?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: What is PAT? I updated the line here and added a sentence for Gitlab repositories: https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R93

CryptoRodeo:
> Let's leave the link out for now. We haven't linked to that document yet from the konflux-ci.dev/docs site ... unless you feel like we should. We can also add the information to the docs here and link to it from my work.

@arewm just to clarify, did you mean the link to the doc you added?

ralphbean (approved):

arewm:
> > Let's leave the link out for now. We haven't linked to that document yet from the konflux-ci.dev/docs site ... unless you feel like we should. We can also add the information to the docs here and link to it from my work.
> 
> @arewm just to clarify, did you mean the link to the doc you added? This one? -> 

Yes, I was questioning the link because I don't think we reference the sample component from the docs repo yet.. but we do have it linked here: https://konflux-ci.dev/docs/advanced-how-tos/building-olm/

This use case is different though. With the other example, we link to it as an example for how to implement the flow. For this case, it felt different because we were linking to it instead of documenting it further here.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Personal Access Token: https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html

I assume that GitLab is supported as long as a component would be able to be onboarded (i.e. the proper secret is created). Is this something that you tested?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Is it to prevent the wrong branch or more precisely to prevent the default branch from being used?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: I think the word "pull" was confusing me. The Containerfile will `COPY` the submodule to the stage so that the content can be built/leveraged for the resulting container.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Ah, by YAML files, you just mean defining the CRs in a yaml and creating them with the CLI?

We need to improve the documentation to include both CLI and UI flows. I don't think that we need to make the call-out in this page.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Ah, interesting. You haven't found a way for a downstream repo to be in GitLab or the submodule? 

(I guess we are talking about this same topic in two threads now).

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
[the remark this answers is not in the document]
arewm: Yeah, that's a valid scenario that I hadn't thought of. I have used submodules mostly as a repo to build out of directly so it didn't matter what changes happened there as much.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Yeah, the title seems better.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: We use `{ProductName}` instead of Konflux and then the docs will render it.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I haven't found a way to deal with private submodules on Gitlab, at this time. Until I can confirm that with a similar test like I ran for the Github repo, I think it's better to leave that out for now. On Gitlab I did get MRs to update my public submodules, but I still need to confirm the private ones.

I'll add it soon after confirming, my Gitlab repo is on Konflux and once it's fully configured and building I can run a test like I did for the private Github submodules.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Good catch! Let me edit it...

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: Should we add a caveat that at least states that GitLab submodules are untested (or at least private GitLab)?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: I might be mistaken, but it looks like I did use `{ProductName}` -> https://github.com/konflux-ci/docs/pull/120/files#diff-8168286ae4fbd44cc1f9e5677aa3a5949ae97f5f391b1466cc253efe6176ccf8R7

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Yeah that's fine, let me add it now....

CryptoRodeo: Caveat added: https://github.com/konflux-ci/docs/pull/120/commits/8fd002b00b89825db0703c22ea161b6f258554df

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: Yeah, I'm just gaining experience defining CRs in a yaml and configuring things in the CLI. This section has been removed

docs/modules/ROOT/pages/how-tos/workflows/git-submodules.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+See the link:https://git-scm.com/docs/gitsubmodules[Git documentation] for more details on configuring git submodules.
+
+=== Onboard the component onto {ProductName}
+
+After creating the downstream repository you can onboard the component. See xref:/how-tos/creating.adoc[Creating applications and components] for more details.
+
+=== Using private repositories as a git submodule:
+
+When using private repositories we encourage that you give the link:https://github.com/apps/red-hat-konflux[Red Hat {ProductName} Github app] access to that private repository.
```
arewm, line 91: This is a GitHub app for our instance and is not a general purpose GitHub app.

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
arewm: I must have gotten confused with an older revision, apologies.

docs/modules/ROOT/pages/how-tos/workflows/git-submodules.adoc
CryptoRodeo: Hmm... should I move this to the downstream docs then?

docs/modules/ROOT/pages/how-tos/worklows/git-submodules.adoc
CryptoRodeo: > Is it to prevent the wrong branch or more precisely to prevent the default branch from being used?

Hmm... it could be both? Maybe someone would prefer another branch besides main, for example a branch a named `version-x` for a specific release

From my own experience it looks like the git submodule defaults to the last commit to that repo, most likely main/master.

docs/modules/ROOT/pages/how-tos/workflows/git-submodules.adoc
CryptoRodeo: I removed this section, see https://github.com/konflux-ci/docs/pull/120/commits/fa1880375b27566d5555d7851d224d82d735605e

arewm (approved):
I think this is good for now. We can continue to iterate on the doc in the future!

Thanks again!

PR #122, opened 2024-08-27
Add customization examples

Hyperkid123:
Based on a conversation with the Konflux engineers, I've added a couple of examples to the docs which might help others in the future.

### Changes
- added an example of how to add a task that generates TA and how to consume it
- added an example of how to add a task with sidecar containers

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
@@ -176,6 +176,59 @@ An example of a custom task added to the pipeline that sends a slack notificatio
 
 Custom Tasks may need access to data from other Tasks. However, in order to not break the chain of trust in a build Pipeline, there are restrictions in modifying such data. For example, a custom Task should not be allowed to modify the component's source code. If you are using the Enterprise Contract (EC) to verify your builds, introducing a custom Task may violate the link:https://enterprisecontract.dev/docs/ec-policies/release_policy.html#trusted_task__trusted[Trusted Tasks] rule. See xref:advanced-how-tos/using-trusted-artifacts.adoc[Trusted Artifacts] for how to safely allow share data between Tasks.
 
+== Extending the build pipeline with your own tasks using Trusted Artifacts
+
+If you are using the Enterprise Contract (EC) to verify your builds, you can extend the build pipeline with your own tasks using Trusted Artifacts. Trusted Artifacts allow you to share data between Tasks in a secure way. For example, the `task-git-clone` task, has Trusted Artifacts version called `task-git-clone-oci-ta`. 
+[NOTE]
+====
+You can find more tasks link:https://quay.io/organization/konflux-ci[here] and their build definitions link:https://github.com/konflux-ci/build-definitions/tree/main/task[here]. Task generating Trusted Artifacts have `-oci-ta` suffix.
+====
+
+.Adding oci-ta task to pipeline 
+
+In your Pipeline definition, add a new task to the `tasks` section.
+
+[source,yaml]
+--
+spec:
+  tasks:
+    # ...
+    - name: clone-repository-oci-ta
+      params:
+      - name: url
+        value: $(params.git-url)
+      - name: revision
+        value: $(params.revision)
+      - name: ociStorage
+        # needs to be unique storage name within the pipeline definition
+        value: $(params.output-image).git
+      runAfter:
+      - init
+      taskRef:
+        params:
+        - name: name
+          value: git-clone-oci-ta
+        - name: bundle
+          value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta@sha256:0f4360ce144d46171ebd2e8f4d4575539a0600e02208ba5fc9beeb2c27ddfd4c
+        - name: kind
+          value: task
+        resolver: bundles
+      workspaces:
+      # use the git-auth workspace for credentials
+      - name: basic-auth
+        workspace: git-auth  
+--
+
+The trusted tasks generate `SOURCE_ARTIFACT` result. This result can be used by other tasks in the pipeline.
+
+.Using `SOURCE_ARTIFACT` in another task
+
+In order to use the `SOURCE_ARTIFACT` in another task, the artifact has to be added as a parameter to the task. The artifact itself is then loaded to a volume accessible only to the task.
+
+You can check link:https://github.com/RedHatInsights/chrome-service-backend/blob/main/.tekton/chrome-service-pull-request.yaml[pipeline] of using the `SOURCE_ARTIFACT` in the pipeline. Look for the `validate-schemas` task.
+
+The task has access to the repository source code, while maintaining the security of the build pipeline and following policies set by EC. This is a common use case for running unit tests, validating the source code, or running a linter.
+
 == Exchanging the build pipeline build task with higher memory limits
 
 The `buildah` task, which builds components from a Dockerfile, has a memory limit of 4 GB. To build components with memory requirements greater than 4 GB, use the following tasks:
--- a/docs/modules/ROOT/pages/how-tos/_nav.adoc
+++ b/docs/modules/ROOT/pages/how-tos/_nav.adoc
@@ -15,6 +15,7 @@
 ** xref:how-tos/testing/index.adoc[Testing your components and applications]
 *** xref:how-tos/testing/build/index.adoc[Build-time tests]
 **** xref:how-tos/testing/build/snyk.adoc[Enabling Snyk]
+**** xref:how-tos/testing/build/sidecars.adoc[Enabling sidecar containers]
 *** xref:how-tos/testing/integration/index.adoc[Integration tests]
 **** xref:how-tos/testing/integration/adding.adoc[Adding an integration test]
 **** xref:how-tos/testing/integration/creating.adoc[Creating a custom integration test]
--- a/docs/modules/ROOT/pages/how-tos/testing/build/sidecars.adoc
+++ b/docs/modules/ROOT/pages/how-tos/testing/build/sidecars.adoc
@@ -1,0 +1,47 @@
+= Enabling sidecar containers
+
+You can add sidecar containers to the build pipeline to run additional tasks alongside the main container. For example, you can add a sidecar container to run a database to enable your unit tests to run against a real database.
+
+The following example demonstrates how to add a sidecar database container to a build pipeline.
+
+[source,yaml]
+--
+spec:
+  tasks:
+  # ...
+   - name: test-db
+      taskSpec:
+        # define the sidecar containers
+        sidecars:
+        - image: quay.io/sclorg/postgresql-15-c9s:latest
+          name: test-db-in-sidecar
+          env:
+          # Configure the env variables required for tests to connect to the database. The variables are specific to the image used.
+            - name: POSTGRESQL_DATABASE
+              value: db-name
+            - name: POSTGRESQL_USER
+              value: db-user
+            - name: POSTGRESQL_PASSWORD
+              value: db-password
+        steps:
+          # make sure to setup the steps. You might need workspaces, artifacts, volumes, secrets, etc. to have the environment ready for the tests
+          - image: registry.access.redhat.com/ubi8/minimal
+            name: test-db-step
+            # Set the environment required for the test-db-step step
+            env:
+              - name: PGSQL_USER
+                value: db-user
+              - name: PGSQL_PASSWORD
+                value: db-password
+              - name: PGSQL_HOSTNAME
+                value: localhost
+              - name: PGSQL_PORT
+                value: 5432
+              - name: PGSQL_DATABASE
+                value: db-name
+            script: |
+              #!/bin/bash
+              # ...
+--
+
+To view a full example using sidecar containers, see the link:https://github.com/RedHatInsights/chrome-service-backend/blob/main/.tekton/chrome-service-pull-request.yaml[here]. Look for `run-unit-tests` task.


docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+          value: git-clone-oci-ta
+        - name: bundle
+          value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta@sha256:0f4360ce144d46171ebd2e8f4d4575539a0600e02208ba5fc9beeb2c27ddfd4c
+        - name: kind
+          value: task
+        resolver: bundles
+      workspaces:
+      # use the git-auth workspace for credentials
+      - name: basic-auth
+        workspace: git-auth
```
arewm, line 219: This looks like the definition for the default git-clone task. Are you suggesting that users add this or are you just using this as an example to show the different parts? If the latter, I think we should use a different pattern.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+    # ...
+    - name: clone-repository-oci-ta
+      params:
+      - name: url
+        value: $(params.git-url)
+      - name: revision
+        value: $(params.revision)
+      - name: ociStorage
+        # needs to be unique storage name
+        value: $(params.output-image).git
```
arewm, line 204: It needs to be globally unique within the pipeline definition, righ?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+
+[source,yaml]
+--
+spec:
+  tasks:
+    # ...
+    - name: task-using-trusted-artifact
+      # make sure the task runs after the artifact is created
+      runAfter:
+        - clone-repository-oci-ta
```
arewm, line 234: Is this strictly necessary? Tekton should be able to order the TaskRuns (but the visualization may be better with the `runAfter`s defined.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+      params:
+        # store the trusted artifact
```
arewm, line 236: This implies that we are storing the artifact here, but we are not. We are resolving the OCI artifact that has been previously stored, right?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+          # set the working directory to value from previous step
+          workingDir: /var/workdir/source
+          # Use image that suites your use case
+          image: registry.access.redhat.com/ubi8/minimal
+          securityContext:
+          # If the task step needs write access to the volume, set the runAsUser to 0 (root). 
+            runAsUser: 0
+          script: |
+            #!/bin/bash
+            # Execute the task
```
arewm, line 270: This ends up being a lot of rendered yaml content. Would it be better to just reference an example in _some_ repository that has the content annotated, to reference some annotated "sample" task definition that has the required steps, or to just have smaller snippets of code isolated to the specific requirements?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -239,6 +336,90 @@ The task update PR will trigger the on-pull-request pipeline, testing the change
...
+          - description: The Trusted Artifact URI pointing to the artifact with the application source code.
+            name: SOURCE_ARTIFACT
+            type: string
+        volumes:
+          - name: workdir
+            emptyDir: {}
+        stepTemplate:
+          volumeMounts:
+            - mountPath: /var/workdir
+              name: workdir
```
arewm, line 379: This part is irrelevant to the sidecar that is added, right? It is just the "normal task definition"?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -239,6 +336,90 @@ The task update PR will trigger the on-pull-request pipeline, testing the change
 
 By default, all pipelines push the images to a local repository that is set up as a part of installation. Ths registry address is registry-service.kind-registry:5001. It is not mandatory to use this local repo, so if you want to use your own Quay repo to control user permissions, you can do this by following link:https://github.com/konflux-ci/konflux-ci/blob/main/docs/quay.md#configuring-a-push-secret-for-the-build-pipeline[the instructions] for configuring a push secret for the build piepline.
 
+== Adding sidecar containers to the build pipeline
+
+You can add sidecar containers to the build pipeline to run additional tasks alongside the main container. For example, you can add a sidecar container to run a database to enable your unit tests to run against a real database.
+
+In the example, we add a sidecar container to task to run a PostgreSQL database. Such sidecar can be used to test database seeding, migrations, etc.
```
arewm, line 343: Is the primary use case for this unit tests? If so, should we add this to the testing pages?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -239,6 +336,90 @@ The task update PR will trigger the on-pull-request pipeline, testing the change
...
+            - name: DATABASE_URL
+              value: "postgresql://$(params.PGSQL_USER):$(params.PGSQL_PASSWORD)@localhost:$(params.PGSQL_PORT)/$(params.PGSQL_DATABASE)"
+            - name: DATABASE_SSL
+              value: "false"
+        steps:
+          - name: use-trusted-artifact
+            image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:8391272c4e5011120e9e7fee2c1f339e9405366110bf239dadcbc21e953ce099
+            args:
+              - use
+              - $(params.SOURCE_ARTIFACT)=/var/workdir
```
arewm, line 402: I guess that this is just needed if your sidecar content is pulled from the git source? This code snippet isn't in the TA section so we need to be careful about the actual content we include without explaining it further.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -239,6 +336,90 @@ The task update PR will trigger the on-pull-request pipeline, testing the change
...
+            type: string
+        volumes:
+          - name: workdir
+            emptyDir: {}
+        stepTemplate:
+          volumeMounts:
+            - mountPath: /var/workdir
+              name: workdir
+        # define the sidecar containers
+        sidecars:
```
arewm, line 381: The content of this feels too specific again which results in the snippet being too large.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
...
+            emptyDir: {}
+
+        stepTemplate:
+          volumeMounts:
+            - mountPath: /var/workdir
+              name: workdir
+        steps:
+        # Add the trusted artifact to the task volume
+        - name: use-trusted-artifact
+          image: quay.io/redhat-appstudio/build-trusted-artifacts:latest@sha256:8391272c4e5011120e9e7fee2c1f339e9405366110bf239dadcbc21e953ce099
```
arewm, line 256: We should use pullspecs from quay.io/konflux-ci wherever possible.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -176,6 +176,103 @@ An example of a custom task added to the pipeline that sends a slack notificatio
 
 Custom Tasks may need access to data from other Tasks. However, in order to not break the chain of trust in a build Pipeline, there are restrictions in modifying such data. For example, a custom Task should not be allowed to modify the component's source code. If you are using the Enterprise Contract (EC) to verify your builds, introducing a custom Task may violate the link:https://enterprisecontract.dev/docs/ec-policies/release_policy.html#trusted_task__trusted[Trusted Tasks] rule. See xref:advanced-how-tos/using-trusted-artifacts.adoc[Trusted Artifacts] for how to safely allow share data between Tasks.
 
+== Extending the build pipeline with your own tasks using Trusted Artifacts
+
+If you are using the Enterprise Contract (EC) to verify your builds, you can extend the build pipeline with your own tasks using Trusted Artifacts. Trusted Artifacts allow you to share data between Tasks in a secure way. For example, the `task-git-clone` task, has Trusted Artifacts version called `task-git-clone-oci-ta`. 
+[NOTE]
+====
+You can find more tasks link:https://quay.io/organization/konflux-ci[here]. Task generating Trusted Artifacts have `-oci-ta` suffix.
```
arewm, line 184: Should we link to quay or to the build definitions?

arewm (changes requested):
Do you think that it would make adding content like this easier if we split up customizing the build pipeline more? For example, we can have a generic customization and then a separate one which is specific to the OCI trusted artifacts?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: If we want to use the TA for custom tasks to comply with the EC, we have to use the `quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta`. We had to add it as a new task.

It is practically the same as the default task. But the default task uses workspaces which means custom tasks will breach the EC.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: Yes. I'll make the comment more specific

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: I wanted to make it obvious and follow the pattern of the generated tekton config provided by Konflux.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: Correct. Poor choice of words.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: We have a sample here: https://github.com/RedHatInsights/chrome-service-backend/blob/main/.tekton/chrome-service-pull-request.yaml#L476-L514

If you prefer linking to an existing example I ok with that.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: Correct. I can just leave the sidecar config. And also link to an existing example if you would prefer it.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: > Is the primary use case for this unit tests?

Correct.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: It is yeah. The main purpose of these examples is to show how to extend pipelines with custom tasks and comply with the EC. But I can see your point.  Maybe instead of these full examples, it should be made very obvious that pipeline customizations need a different approach, especially if the builds are using "external" policies not defined by the project itself.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: We can link to an existing example if you'd like.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: Is there an equivalent spec in the konflux-ci org? I did not find any.

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
Hyperkid123: We could link here: https://github.com/konflux-ci/build-definitions/tree/main/task, these do not have the image repository links. So one does not really know form where to pull the images. So right now, probably link to both?

Hyperkid123:
@arewm I've updated the PR. Removed some of the YAML and moved the sidecar example to the testing section. I've also added links to existing pipelines so folks can get the full context. I am not linking directly to the task's code lines. They will likely change in the future and the links might lead to a different spec.

Let me know what you think 🙂 .

PR #127, opened 2024-09-05
Add some frequently used commands

ralphbean:
These are commands that I keep in my shell history and find myself frequently using to navigate around, and sharing the commands with others to do the same.

~~Here, I put them at the *top* of the doc in the hopes that it can be a quick reference for would-be power users.~~


arewm (commented):
I think it still makes sense to add these lower in the page as we miss the context of what we are exploring and why we need to parse it like this.

Even if we include it lower on the page, the content can be directly referenced using page anchors.

ralphbean:
OK, agreed. Updated!

arewm (approved):

PR #128, opened 2024-09-06
Make a note about confirming RP matches

ralphbean:


docs/modules/ROOT/pages/advanced-how-tos/releasing/create-release-plan.adoc
```diff
@@ -61,3 +61,4 @@ $ oc apply -f ReleasePlan.yaml -n dev
 
 . In the {ProductName} UI, select the *Release services* > *Release plan* tab.
 . Review the Release plan object that you just added. Using the Release plan tab, you can update or delete the selected Release plan object.
+. Confirm that your ReleasePlan is successfully *matched* to a ReleasePlanAdmission by looking at your ReleasePlans in the "Releases" section of the UI in the menu on the left hand side or by inspecting the `status` of the ReleasePlan resource in the API.
```
arewm, line 64: We have some information on matching with the RPA: https://github.com/konflux-ci/docs/blob/main/docs/modules/ROOT/pages/advanced-how-tos/releasing/create-release-plan-admission.adoc?plain=1#L68

Should we have information in both or does it make sense to have it only the RP or RPA?

ralphbean: If we were to have it in only one place, I think the `RP` is the right place. The RP is the thing that the user tries to use when they create a `Release` and is therefore the natural place to look for a misconfiguration status.

That said, I think it's fine having it in both places.

arewm: Having it in both places works. Can we make the text consistent?

ralphbean: Yes, done.

PR #129, opened 2024-09-06
Try adding an index of build tasks

ralphbean:
This is based on https://github.com/konflux-ci/build-definitions/pull/1381

Related: https://issues.redhat.com/browse/KONFLUX-1229

### DIFF
--- a/docs/modules/ROOT/pages/tasks/build.adoc
+++ b/docs/modules/ROOT/pages/tasks/build.adoc
@@ -4,130 +4,130 @@ These tasks, defined in the https://github.com/konflux-ci/build-definitions[buil
 You can use references to these tasks when xref:/how-tos/configuring/customizing-the-build.adoc#configuring-timeouts[customizing your build pipeline].
 
 
-[[acs-deploy-check]]acs-deploy-check (0.1):: Policy check a deployment with StackRox/RHACS This tasks allows you to check a deployment against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
+[[acs-deploy-check-0.1]]acs-deploy-check (0.1):: Policy check a deployment with StackRox/RHACS This tasks allows you to check a deployment against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
 
-[[acs-image-check]]acs-image-check (0.1):: Policy check an image with StackRox/RHACS This tasks allows you to check an image against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
+[[acs-image-check-0.1]]acs-image-check (0.1):: Policy check an image with StackRox/RHACS This tasks allows you to check an image against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
 
-[[acs-image-scan]]acs-image-scan (0.1):: Policy check an image with StackRox/RHACS This tasks allows you to check an image against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
+[[acs-image-scan-0.1]]acs-image-scan (0.1):: Policy check an image with StackRox/RHACS This tasks allows you to check an image against build-time policies and apply enforcement to fail builds. It's a companion to the stackrox-image-scan task, which returns full vulnerability scan results for an image. See also:
 
-[[apply-tags]]apply-tags (0.1):: Applies additional tags to the built image. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/apply-tags/0.1/README.md[README.md]
+[[apply-tags-0.1]]apply-tags (0.1):: Applies additional tags to the built image. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/apply-tags/0.1/README.md[README.md]
 
-[[buildah]]buildah (0.1):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/buildah/0.1/README.md[README.md]
+[[buildah-0.1]]buildah (0.1):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/buildah/0.1/README.md[README.md]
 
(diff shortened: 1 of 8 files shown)


ralphbean:
I'm posting this here (prematurely) just to see what it looks like in the PR preview build.

ralphbean:
Well, that's something: https://pr-129--konflux-docs.netlify.app/tasks/

docs/modules/ROOT/pages/tasks/build.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+[[buildah-remote-0.2]]buildah-remote (0.2):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-remote/0.2/README.md[README.md] https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-remote/0.2/MIGRATION.md[MIGRATION.md]
+
+[[buildah-remote-oci-ta-0.1]]buildah-remote-oci-ta (0.1):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also:
+
+[[buildah-remote-oci-ta-0.2]]buildah-remote-oci-ta (0.2):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-remote-oci-ta/0.2/README.md[README.md] https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-remote-oci-ta/0.2/MIGRATION.md[MIGRATION.md]
+
+[[buildah-rhtap-0.1]]buildah-rhtap (0.1):: Buildah task builds source code into a container image and pushes the image into container registry using buildah tool. See also:
```
arewm, line 31: Did we decide that we will not filter the list of tasks to remove ones like this?

arewm, line 31: `See also:` is printed even though there is nothing to see.

docs/modules/ROOT/pages/tasks/build.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+[[tkn-bundle-0.1]]tkn-bundle (0.1):: Creates and pushes a Tekton bundle containing the specified Tekton YAML files. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/tkn-bundle/0.1/README.md[README.md]
+
+[[update-deployment-0.1]]update-deployment (0.1):: Task to update deployment with newly built image in gitops repository. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/update-deployment/0.1/README.md[README.md]
+
+[[update-infra-deployments-0.1]]update-infra-deployments (0.1):: Clones redhat-appstudio/infra-deployments repository, runs script in 'SCRIPT' parameter, generates pull-request for redhat-appstudio/infra-deployments repository. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/update-infra-deployments/0.1/README.md[README.md]
+
+[[upload-sbom-to-trustification-0.1]]upload-sbom-to-trustification (0.1):: Upload an SBOM file to [Trustification] using the [BOMbastic] API. See also: https://github.com/konflux-ci/build-definitions/tree/main/task/upload-sbom-to-trustification/0.1/README.md[README.md]
```
arewm, line 129: The `[Trustification]` and `[BOMbastic]` references are hyperlinks in the README but are not rendered properly here.

hack/generate-tekton-index.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo >> "$WORKDIR/tasks.adoc"
+    echo -n "[[$task_name-${task_version}]]$task_name " >> "$WORKDIR/tasks.adoc"
+
+    if [ "${task_version}" == "" ]; then
+        task_version=$(yq '.metadata.labels."app.kubernetes.io/version" // "undefined"' "$task_dir/$task_name.yaml")
+    fi
+
+    echo -n "($task_version):: " >> "$WORKDIR/tasks.adoc"
+
+    description=$(yq '.spec.description | split("\n") | .[0]' "$task_dir/$task_name.yaml")
```
arewm, line 59: I implemented something similar in https://github.com/konflux-ci/build-definitions/pull/1397/files#diff-78a47706e6b9aca3579f93bc34adac628f9c102de7257a2fe14939130786a807R123. Is this a reason that we should do this within yq or is it just convenience?

zregvart:
The EC team [had used](https://github.com/enterprise-contract/ec-cli/tree/b2ddd3deeef6c3aa88bec1ada230f28b2dcdd3ce/tekton-task-antora-extension) an [Antora extension](https://docs.antora.org/antora/latest/extend/extensions/) that provided the deserialized Task object to a [Handlebars template](https://github.com/enterprise-contract/ec-cli/blob/b2ddd3deeef6c3aa88bec1ada230f28b2dcdd3ce/docs/template/tekton-task.hbs) for the Task page and for [the navigation](https://github.com/enterprise-contract/ec-cli/blob/b2ddd3deeef6c3aa88bec1ada230f28b2dcdd3ce/docs/modules/ROOT/partials/tasks_nav.adoc?plain=1). This has the benefit of pointing to the YAML Task files and generating the documentation from them.

We decided that a better approach would be to generate the Asciidoc files at [build time](https://github.com/enterprise-contract/ec-cli/tree/e260a859e2906254735902b0af6b808fab01de14/internal/documentation/asciidoc/tekton), and [mandate](https://github.com/enterprise-contract/ec-cli/blob/b2ddd3deeef6c3aa88bec1ada230f28b2dcdd3ce/.github/workflows/checks.yaml#L63-L74) that they're committed in git. We decided not to write shell scripts for the templating.

The drawback of the approach taken here is that it ties the layout/content of build-definitions repository to the docs repository here and with that a change in the build-definitions could cause the documentation not to build or build incorrectly. I think the approach of having the Asciidoc files in the build-definitions repository could be something that we consider.

hack/generate-tekton-index.sh
ralphbean: Oh, just convenience I think. `head` works too.

ralphbean:
Ack. I am uncertain what's going to happen with antora here w.r.t. @MartinBasti's ongoing investigation. Maybe we should put this on hold until that settles; then, revisit here.

docs/modules/ROOT/pages/tasks/build.adoc
ralphbean: Good catch! Let's filter it.

ralphbean:
Related prior work: https://github.com/konflux-ci/build-definitions/pull/756

ralphbean:
I think we still want something like this done, but I want to drop this attempt. Someone else can do it better.

PR #136, opened 2024-09-20
Document how to configure rpm prefetch

ralphbean:
This required adjusting language at the top of the document about what is supported.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
@@ -321,7 +321,7 @@ contentOrigin:
 . Copy any necessary yum/dnf repo files into your git repository. If you are using a fedora rawhide base image, that looks like:
 
 +
-[source]
+[source,console]
 ----
 $ BASE_IMAGE=quay.io/fedora/fedora:rawhide
 $ podman run -it $BASE_IMAGE cat /etc/yum.repos.d/fedora.repo > fedora.repo
@@ -333,7 +333,7 @@ NOTE: For every repository defined in your set of repo files, make sure to add t
 . Run the following command to resolve your `rpms.in.yaml` file and produce a `rpms.lock.yaml` file.
 
 +
-[source]
+[source,console]
 ----
 $ BASE_IMAGE=quay.io/fedora/fedora:rawhide
 $ rpm-lockfile-prototype --image $BASE_IMAGE rpms.in.yaml <1>
--- a/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
@@ -293,6 +293,7 @@ spec:
 * From the {ProductName} *Applications* view, go to *Activity > Pipeline runs*.
 ** Go to the pipeline run with *Build* in the *Type* column and confirm that the `prefetch-dependencies` stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.
 * From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
+//include::../partials/con_hermetic_verification.adoc[]
 
 == Enabling prefetch builds for `rpm`
 
@@ -314,8 +315,8 @@ packages: [nethack] <1>
 contentOrigin:
   repofiles: ["./fedora.repo"] <2>
 ----
-<1> The `*packages*` list is the list of packages you want to install in your Container. Only include first-order dependencies here, i.e. the packages you want to install.
-<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the Cachi2 where to find your rpm and its dependencies.
+<1> The `*packages*` list is the list of packages you want to install in your Container. You don't have to declare transitive dependencies here. The rpm-lockfile-prototype tool will resolve them for you.
+<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the tooling where to find your rpm and its dependencies.
 
 . Copy any necessary yum/dnf repo files into your git repository. If you are using a fedora base image, that looks like:
 
@@ -371,6 +372,7 @@ spec:
 * From the {ProductName} *Applications* view, go to *Activity > Pipeline runs*.
 ** Go to the pipeline run with *Build* in the *Type* column and confirm that the `prefetch-dependencies` stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.
 * From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
+//include::../partials/con_hermetic_verification.adoc[]
 
 .Troubleshooting
 If your build fails, be sure to look at your logs:
--- a/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
@@ -349,12 +349,11 @@ spec:
 ----
 <1> The `*prefetch-input*` parameter specifies the path to the directory that has the lockfile. In this example, the `.` indicates that the rpm lockfile is in the repository root. Additionally, if you have multiple directories, you can provide the path to those directories in the JSON array format. For example, `[{"path": ".", "type": "rpm"}, {"path": "subpath/to/the/other/directory", "type": "rpm"}]`.
 
-. Additionally, pass an extra parameter to the `prefetch-dependencies` task to indicate that "dev package managers" should be enabled.
+. Additionally, pass an extra parameter to the `prefetch-dependencies` task in the `.spec.pipelineSpec.tasks` section to indicate that "dev package managers" should be enabled.
 
 +
 [source,yaml]
 ----
-spec:
     tasks:
         -   ...
         -   name: prefetch-dependencies
--- a/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
@@ -4,8 +4,8 @@ In {ProductName}, you can run a hermetic build by restricting network access to 
 
 For every build, Cachi2 generates a software bill of materials (SBOM) where all dependencies are properly declared and pinned to specific versions. Also, Cachi2 ensures that arbitrary code is never executed during the prefetch, meaning, for example, that the build doesn’t pull any undeclared dependencies. Such measures result in very accurate SBOMs and improve the build reproducibility. For more information about SBOMs, see xref:/how-tos/metadata/sboms.adoc[Inspecting SBOMs].
 
-[#supported-languages]
-.Supported languages
+[#available-languages]
+.Available languages
 
 [cols="1,1"]
 |===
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+. Create a `rpms.in.yaml` file in your git repository, with the following contents:
+
++
+[source,yaml]
+----
+packages: [nethack] <1>
+contentOrigin:
+  repofiles: ["./fedora.repo"] <2>
+----
+<1> The `*packages*` list is the list of packages you want to install in your Container. Only include first-order dependencies here, i.e. the packages you want to install.
```
chmeliik, line 317: I'm not sure if readers will understand "first-order dependencies." Maybe a note like this would be sufficient: "You don't have to declare transitive dependencies here, the `rpm-lockfile-prototype` tool will resolve them for you."

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the Cachi2 where to find your rpm and its dependencies.
```
chmeliik, line 318:
```suggestion
<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the tooling where to find your rpm and its dependencies.
```
More generic might be better. Currently, only `rpm-lockfile-prototype` relies on the repo files, cachi2 does not. But that may change in the future.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+<1> You won't find `dev-package-mangers` as a param on the `prefetch-dependencies` task. You have to add it, and set it to true. This is because Cachi2 hasn't declared stable support for rpm lockfile processing yet. It's new technology and the link:https://github.com/rpm-software-management/dnf5/issues/833[conversation] about which way forward in the dnf community is still ongoing.
+
+. Create a pull request by committing your changes to the repository of the component.
+. Review and merge the pull request.
+
+.Verification
+* From the {ProductName} *Applications* view, go to *Activity > Pipeline runs*.
+** Go to the pipeline run with *Build* in the *Type* column and confirm that the `prefetch-dependencies` stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.
+* From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
+
```
chmeliik, line 374: Was something supposed to follow here?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+----
+<1> The `*prefetch-input*` parameter specifies the path to the directory that has the lockfile. In this example, the `.` indicates that the rpm lockfile is in the repository root. Additionally, if you have multiple directories, you can provide the path to those directories in the JSON array format. For example, `[{"path": ".", "type": "rpm"}, {"path": "subpath/to/the/other/directory", "type": "rpm"}]`.
+
+. Additionally, pass an extra parameter to the `prefetch-dependencies` task to indicate that "dev package managers" should be enabled.
+
++
+[source,yaml]
+----
+spec:
+    tasks:
```
chmeliik, line 357:
```suggestion
spec:
    pipelineSpec:
        tasks:
```
Setting prefetch-input is done in the PipelineRun, while adding dev-package-managers is done in the Pipeline. This way, at least both the snippets look like a PipelineRun

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+<1> The `*packages*` list is the list of packages you want to install in your Container. Only include first-order dependencies here, i.e. the packages you want to install.
+<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the Cachi2 where to find your rpm and its dependencies.
+
+. Copy any necessary yum/dnf repo files into your git repository. If you are using a fedora base image, that looks like:
+
++
+[source]
+----
+$ cp /etc/yum.repos.d/fedora.repo .
+----
```
chmeliik, line 326: Should we also warn about enabling source repos? A snippet from an internal doc:

> ⚠️ For every repository defined in your set of repo files, make sure to add the corresponding sources repo (or make sure to enable them, if they’re already present). Otherwise, the lockfile generator will not include any SRPMs in your lockfile, cachi2 won’t download any SRPMs and the source container for your build will be incomplete.

> This is the resulting repo file for the example build: [ubi.repo](https://github.com/chmeliik/build-tasks-dockerfiles/blob/konflux-hermetic/task-toolset/ubi.repo). Note that, unlike the original file in the ubi-minimal container image, it sets enabled = 1 for the -sources repos.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
 ** Go to the pipeline run with *Build* in the *Type* column and confirm that the `prefetch-dependencies` stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.
 * From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
 
+== Enabling prefetch builds for `rpm`
+
+Cachi2 has a dev-preview package manager capable of fetching `rpm` dependencies. This requires the use of a pair of `rpms.in.yaml` and `rpms.lock.yaml` files to be committed to your repository. You write a `rpms.in.yaml` file and the link:https://github.com/konflux-ci/rpm-lockfile-prototype?tab=readme-ov-file#what-is-this[rpm-lockfile-prototype] CLI tool resolves that to produce a `rpms.lock.yaml` file. Cachi2 fetches those specific rpms and enables your build to install them without network access.
+
+.Prerequisites
+* You have an up-to-date installation of link:https://github.com/konflux-ci/rpm-lockfile-prototype?tab=readme-ov-file#installation[rpm-lockfile-prototype].
```
arewm, line 302: You also need to run this on a system with available commands ... i.e. a user wouldn't be able to run this on a mac.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+[source,yaml]
+----
+packages: [nethack] <1>
+contentOrigin:
+  repofiles: ["./fedora.repo"] <2>
+----
+<1> The `*packages*` list is the list of packages you want to install in your Container. Only include first-order dependencies here, i.e. the packages you want to install.
+<2> This should be a reference to a repo file, like those found in `/etc/yum.repos.d/`. This tells the Cachi2 where to find your rpm and its dependencies.
+
+. Copy any necessary yum/dnf repo files into your git repository. If you are using a fedora base image, that looks like:
```
arewm, line 320: This is the base image of the container that you are building? Would users know how to select the appropriate file(s) for rhel/UBI from this?

arewm, line 320: Should this be copied from the system that is running the tool or should we instruct users on how to extract the file from the actual parent image that will be used?

(or should the tool be run from the base image which will be used?)

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -289,6 +294,84 @@ spec:
...
+----
+$ cp /etc/yum.repos.d/fedora.repo .
+----
+
+. Run the following command to resolve your `rpms.in.yaml` file and produce a `rpms.lock.yaml` file.
+
++
+[source]
+----
+$ rpm-lockfile-prototype --image YOUR_BASE_IMAGE rpms.in.yaml <1>
```
arewm, line 333: Ah, I assume that this means that the prior base image is indeed that of the container you are building.

arewm (changes requested):
Thanks for adding this!

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: Ah, I was missing

```
//include::../partials/con_hermetic_verification.adoc[]
```

Adding that in now.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: How about the change in a42dcb3a1d0aa6d9c737cc410ba1273abd37ee3f?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: Yes, good call.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: Hm, let's provide a command to extract it from the base image they're working with.

My favorite command for that is `oc image extract ...` though, and I don't know an "easy" alternative with skopeo.

ralphbean: Well, I guess this works:

```
podman run -it registry.redhat.io/ubi9/ubi:latest cat /etc/yum.repos.d/ubi.repo > ubi.repo
```

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
chmeliik: That works :+1:

chmeliik (approved):

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: Updated in 27cdf67309ca42a630f86cf7c96eae3f0045dbde

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
ralphbean: Yes. Clarified in 27cdf67309ca42a630f86cf7c96eae3f0045dbde.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
brunoapimentel: The *.repo files don't really need to be commited to the user's repo, they only need to be there for the lockfile resolution (although it doesn't hurt to commit them). Not sure if we should point that out, or if that'll only make it more confusing.

ralphbean: I am similarly unsure if one is more confusing than the other.

I guess, without more user reactions to go on, I'd assume that committing the file is more explicit and clear. FWIW, it kind of looks like a beefier version of content_sets.yaml (some openstack devs today understood it right away when I tested out explaining it that way).

arewm (approved):

PR #141, opened 2024-10-08
Change to dev-package-managers value being default

arewm:
As this is a parameter for the pipelineSpec, setting the value must be done with the default key instead of the value key.


arewm:
Follow-on from #136.

arewm:
Nevermind. I misread the documentation. I just tried adding the parameter to the pipelinespec.params and not to the pipelinespec.tasks.params

PR #145, opened 2024-10-11
Only suggest to use git branches for build references

arewm:
Configuring components with immutable references like tags and commits can work to build artifacts but it doesn't work when trying to build new commits. Removing these options from the documentation.


dirgim (approved):
LGTM

PR #149, opened 2024-10-14
Update creating-secrets.adoc

ernesgonzalez33:
Remove the "personal" instructions from the access tokens in GitLab to a "project" access token

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/creating-secrets.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/creating-secrets.adoc
@@ -57,10 +57,11 @@ When building content from GitHub, you will need to install an application for P
 
 .Procedure (GitLab)
 
-. In GitLab select your avatar, then select **Edit profile** > **Access Tokens** > **Add new token**.
+. In GitLab click on Settings -> Access Tokens on the left menu of your repository. **NOTE:** if you don't see this option, you need someone with permissions to do so.
+. Then click on `Add new token`.
 . Select the following scopes: `api`, `read_repository`, and `write_repository`.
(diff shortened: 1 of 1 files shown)


arewm (approved):

arewm:
@ernesgonzalez33 , is it possible to have a project access token that has access to more than one repository? Are there any caveats that we need to add?

ernesgonzalez33:
> @ernesgonzalez33 , is it possible to have a project access token that has access to more than one repository? Are there any caveats that we need to add?

@arewm I added more info. WDYT?

docs/modules/ROOT/pages/how-tos/configuring/creating-secrets.adoc
```diff
@@ -93,6 +94,8 @@ Using the PAT authentication requires only the `password` key. The `username` sh
 
 This secret will be used by the build service to perform builds with Pipeline-as-Code.
 
+This Project Access Tokens are scoped to a project, so you cannot use them to access resources from other projects. For more information regarding Project Access Tokens, see the Gitlab https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html[documentation].
```
arewm, line 97: What implications does this have if users want to onboard multiple projects? How can this be accomplished?

ernesgonzalez33: That is already in the docs written just below these lines

arewm: Ah, thanks.

arewm (approved):

PR #151, opened 2024-10-15
Remove unused parameters from custom integration test example.

tmds:
I assume these parameters are not relevant for the example because they are not covered in the rest of the document.

@arewm ptal.


arewm (commented):
I'll defer approval to @dirgim who has more background.

PR #157, opened 2024-10-22
Miscellaneous fixes

chmeliik:
Problems discovered during the Konflux Onboarding Accelerator week in Brno


chmeliik:
Note that the tenant release pipeline is still completely unusable, the `push-to-external-registry.yaml` pipeline only works as a managed release pipeline. A working example for a tenant release pipeline does not exist at the moment

arewm (approved):
Thanks for these changes. Should we remove the pipeline reference to one that _does_ work even if it has different functionality?

chmeliik:
> Thanks for these changes. Should we remove the pipeline reference to one that _does_ work even if it has different functionality?

Per my understanding, no useful example pipeline exists at the moment. It's been requested in https://issues.redhat.com/browse/CLOUDDST-24698

PR #159, opened 2024-10-24
Add note on how to get access to quay on release

mshriver:
Multiple threads in the konflux-users channel trying to figure out why they can't see the `redhat-services-prod` org. 

https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1729768334150819

The context here will change and the text should likely update with:
https://issues.redhat.com/browse/RELEASE-570


arewm:
Thanks for your contributions to improve the documentation. While this is a common issue, I think it is more relevant to an internal deployment of Konflux than it is to the community Konflux project.

PR #170, opened 2024-11-14
Add a reference to how to link secrets to SAs on manually created secrets

Victoremepunto:
closes KFLUXBUGS-1766

### DIFF
--- a/docs/modules/ROOT/pages/troubleshooting/index.adoc
+++ b/docs/modules/ROOT/pages/troubleshooting/index.adoc
@@ -102,6 +102,7 @@ skopeo login --authfile /tmp/auth.json "$should_have_access_to"
 
 Note: works for `dockercfg` and `dockerconfigjson` secrets, not `basic-auth`.
 
+[#check-if-the-secret-is-linked-to-the-service-account]
 === Check if the secret is linked to the `appstudio-pipeline` service account
 
 In order to connect the new secret to pipeline run, you need to link it from service account used by Tekton to run the pipeline run.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/troubleshooting/index.adoc
```diff
@@ -118,7 +119,11 @@ If not, link it to the `appstudio-pipeline` service account with:
 ----
 secret_name=secret-for-my-registry
 
-kubectl secret link appstudio-pipeline "$secret_name"
+# using oc
+oc secrets link appstudio-pipeline "$secret_name"
+
+# using kubectl
+kubectl patch serviceaccount appstudio-pipeline -p '{"secrets": [{"name": "$secret_name"}]}'
```
arewm, line 126: Do we need to be specific about `oc` usage here? I tried to describe everything in terms of `kubectl` within the docs.

Victoremepunto: not really, but I was confused if that was the actual intent of the original author, since `kubectl secret link` doesn't exist, but `oc secrets link` does.

I added the correct kubectl version (which is a bit cumbersome being a patch op., specially compared to the `oc` version)

But I do agree it stands out a bit since it's the only `oc` command. I can remove it then, would that be better then ?

arewm: Ah. Good catch. How about if you switch the two to put kubectl first then oc? The oc command is certainly simpler!

Victoremepunto: done! 

I've put the kubectl first, then left the oc one 

I actually caught a "quotes problem" when giving it a final test too - now it works perfectly

arewm (approved):
Thanks!

PR #175, opened 2024-11-19
Document Cachi2's generic fetcher

kosciCZ:
Cachi2's generic fetcher has recently emerged from the experimental phase and is now officially supported. This PR adds that documentation also to konflux docs.

(Note - the support comes with Cachi2's next release, which should happen today (2024-11-19), so this should be merged after that as to minimize user confusion)

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
+++ b/docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
@@ -23,6 +23,9 @@ For every build, Cachi2 generates a software bill of materials (SBOM) where all 
 
 |xref:rpm[RPM*]
 |`rpm`
+
+|xref:generic[Generic dependencies]
+|`generic fetcher`
 |===
 
 NOTE: To use Yarn as the package manager, see the link:https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#example-yarn[Cachi2 Usage guide].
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -372,6 +375,58 @@ spec:
 . Create a pull request by committing your changes to the repository of the component.
 . Review and merge the pull request.
 
+.Verification
+* From the {ProductName} *Applications* view, go to *Activity > Pipeline runs*.
+** Go to the pipeline run with *Build* in the *Type* column and confirm that the `prefetch-dependencies` stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.
```
arewm, line 380: fetched all configured dependencies?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -372,6 +375,58 @@ spec:
...
+* From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
+//include::../partials/con_hermetic_verification.adoc[]
+
+.Troubleshooting
+If your build fails, be sure to look at your logs:
+
+In {ProductName}, from the *Applications* view, select the application build you want to troubleshoot, then from the resulting *Overview* page, select the *Activity* tab. From there, under *Activity By*, select *Pipeline runs*. From the *Name* column, select the build whose logs you want to check, then from the resulting *Pipeline run details* view, do one of the following:
+
+* Select the *Logs* tab.
+* Alternatively, you can click *build-container*. When the right panel opens, select the *Logs* tab to see a partial view of the log for that build.
```
arewm, line 390: Are we trying to get them to troubleshoot prefetch or builds?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -372,6 +375,58 @@ spec:
...
+
+== [[generic]]Enabling prefetch builds for `Generic dependencies`
+If you need to prefetch arbitrary files for your build, Cachi2 supports `generic fetcher` for that purpose. It uses a custom lockfile named `artifacts.lock.yaml` to achieve this. This file needs to be either commited in the source repository, or explicitly specified as an absolute path. The latter is useful in case you for some reason need the lockfile to be dynamic and committing it to the repository would be problematic. For more information on supported types of artifacts, see link:https://github.com/containerbuildsystem/cachi2/blob/main/docs/generic.md[Cachi2 documentation].
```
arewm, line 393: What does this mean "specified as an absolute path"? Are the absolute paths not within the repo as well?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -372,6 +375,58 @@ spec:
...
+
+.Procedure
```
arewm, line 395: Should we link to the cachi2 documentation page for more information/detail/context con configuring prefetch?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
kosciCZ: Cachi2 really prefers to work on a clean source repo. So in cases when the lockfile cannot be easily present in a clean repository, this option allows to e.g. mount another workspace and provide the lockfile from there, completely outside of the source repository.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
kosciCZ: Possibly if the prefetch task passes, but the build still fails. This section is copied from the above sections that also include these instructions.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
kosciCZ: Changed.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
kosciCZ: Linked Cachi2 docs for creating the lockfile (the generic fetcher's documentation)

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
arewm: I noticed that it was copied once I saw the same content was there after. We can leave it as it is for now.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
arewm: Yes, I think it would be good to document the behavior especially if we anticipate that it is going to be a desired repeated behavior.

Does this behavior technically exist for any of the prefetch tasks? Do you think that it should be disallowed to "trick" the system by performing tasks like clone -> generate lockfile -> prefetch? Does it matter if the content is attached with an additional mount vs. just modifying the source? Is this just the [`CONIG_FILE_CONTENT`](https://github.com/konflux-ci/build-definitions/blob/main/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml#L137-L145) path?

kosciCZ: > Does this behavior technically exist for any of the prefetch tasks? 

No, I don't think so. Cachi2 usually works with established package managers, lockfiles of which are usually in the repository anyway. However, with lockfiles for the generic fetcher, these are hermetic build specific and might not be found in "upstream" repositories. Another usecase is in the PR I linked, where some dependencies are pre-built as part of the pipeline, making the lockfile effectively dynamic.

> Do you think that it should be disallowed to "trick" the system by performing tasks like clone -> generate lockfile -> prefetch? Does it matter if the content is attached with an additional mount vs. just modifying the source? 

I don't it should be disallowed, seems like a valid use-case to me.  Here's the discussion we had about this in [cachi2 repository](https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1833363333). The idea is that cachi2 can (and by some opinions should) enforce a clean source repo, however that can still be worked around by just committing the file locally in the pipeline. So this option is a "clean" way of supporting this use case. 

> Is this just the [CONIG_FILE_CONTENT](https://github.com/konflux-ci/build-definitions/blob/main/task/prefetch-dependencies-oci-ta/0.1/prefetch-dependencies-oci-ta.yaml#L137-L145) path?
Not sure what you mean by this, but if you're suggesting the lockfile should be placed there in the cases where it is not commited to the repository, I personally don't see an issue with that.

arewm (approved):
Thanks. We can follow up with additional PRs to improve the information if needed.

PR #178, opened 2024-11-22
docs(STONEINTG-1065): Add instructions for rerunning all Integration tests

dheerajodha:


arewm (commented):
I feel like there is a lot of duplication in this page. Can it be restructured to the supported values for `test.appstudio.openshift.io/run` with what those values mean?

dirgim (approved):
LGTM

arewm (approved):
Thanks

PR #173, opened 2024-11-18
Keep remote pipelines up to date documentation

fabriziosta:

PR #179, opened 2024-11-22
Document how to add custom tasks to TA pipelines

chmeliik:


docs/modules/ROOT/pages/how-tos/configuring/adding-custom-tasks.adoc
```diff
@@ -0,0 +1,146 @@
...
+With one major restriction: they *cannot modify your source repo or the prefetched content* before the build.
+Doing so would violate the link:https://enterprisecontract.dev/docs/ec-policies/release_policy.html#trusted_task__trusted[Trusted Tasks] Enterprise Contract rule.
+
+What this means in Trusted Artifacts terms:
+
+* To operate on your repo, your Task must declare a `SOURCE_ARTIFACT` parameter and use a special step to download this artifact
+  (refer to the ready-to-use examples below).
+* Your Task must *not* return a modified `SOURCE_ARTIFACT`.
+  It's ok if the task modifies the source repo as it executes, but the modifications will be discarded.
+  The build will use the original, unmodified `SOURCE_ARTIFACT`.
```
arewm, line 15: nit: 

* It is fine to generate the modified source artifact but that shouldn't be used in any future task that needs to be trusted.
* This is true for _any_ artifacts consumed (source or cachi2) by trusted tasks (prefetch or build)

docs/modules/ROOT/pages/how-tos/configuring/adding-custom-tasks.adoc
```diff
@@ -0,0 +1,146 @@
...
+.Prerequisites:
+
+. Your pipeline uses xref:advanced-how-tos/using-trusted-artifacts.adoc[Trusted Artifacts].
+
+.Procedure:
+
+. Implement the custom Task in the `.tekton/tasks/` directory in your repo.
++
+NOTE: The presence of the Task in `.tekton/**` is enough to make it available for use in your Pipelines without any extra configuration;
+    see link:https://pipelinesascode.com/docs/guide/resolver/[Pipelines-as-Code resolver].
```
arewm, line 29: Another PR is proposing a different structure (for remote resolvers). Should we address the potential confusion?

https://github.com/konflux-ci/docs/pull/173/files#diff-7de75443c634564136fd1db89ef54619c2483dd2be6d8063cd0ffd764e96476dR73-R82

docs/modules/ROOT/pages/how-tos/configuring/adding-custom-tasks.adoc
```diff
@@ -0,0 +1,146 @@
...
+* link:https://tekton.dev/docs/pipelines/tasks/[Tekton Task docs]
+* link:https://github.com/konflux-ci/build-definitions/tree/main/task[Existing Konflux tasks]
+
+== Examples
+
+=== Dynamically set the ARGs and LABELs for the container build
+
+Implement the task in e.g. `.tekton/tasks/get-buildah-args.yaml`:
+
+[source,yaml]
```
arewm, line 43: We have a repository where we can create some sample pipelines/tasks: https://github.com/konflux-ci/pipeline-samples

What do you think about propose this task there instead so that readers won't have to extract it out of the documentation?

docs/modules/ROOT/pages/how-tos/configuring/adding-custom-tasks.adoc
```diff
@@ -0,0 +1,146 @@
+= Adding custom tasks
+
+If your pipeline uses xref:advanced-how-tos/using-trusted-artifacts.adoc[Trusted Artifacts],
```
arewm, line 3: We only talk about TA pipelines here. Can we have a section at the top describing the overall process and then two sections later discussing the uniqueness for TA and PVC pipeline tasks?

docs/modules/ROOT/pages/how-tos/configuring/customizing-the-build.adoc
```diff
@@ -97,44 +96,14 @@ NOTE: This example shows a separate Pipeline file, but your pipeline may also be
...
-  taskRef:
-    params:
-    - name: name
-      value: example-task  # metadata.name field of the Task
-    - name: bundle
-      value: quay.io/tekton-bundle-catalog/example-task-bundle:1.0
-      # For more details on tekton bundles, refer to https://tekton.dev/docs/pipelines/pipelines/#tekton-bundles
-    - name: kind
-      value: task
-    resolver: bundles
```
arewm, line 128: This information is relevant for remote resolvers (along with the undocumented git resolver). Should we include some information/reference to these?

docs/modules/ROOT/pages/how-tos/configuring/adding-custom-tasks.adoc
chmeliik: Hmm, interesting. They are different use cases, but we probably do need to differentiate between them somewhere

That PR: defining a complete "remote" Pipeline (will be referenced from other repos)
This PR: adding a custom Task to a non-remote Pipeline

arewm: Yeah. I don't feel like this PR needs to include both local and remote pipelines/tasks, but maybe just an acknowledgement the the other exists as well?

Should we just keep this PR scoped down to creating your own custom tasks? Then there can be a separate location where we talk about using remote tasks? These pages can be located next to each other.

chmeliik:
Marking as draft until I have time to rework this

PR #182, opened 2024-11-26
Drop the common prefix of ImageRepository names

chmeliik:
In short: the common prefix of ImageRepository names is too long, which triggers https://issues.redhat.com/browse/SRVKP-6798

Drop the prefix.


arewm (approved):
@ifireball , does this look good to you as well?

ifireball (approved):
yeah, lgtm. I probably need to fix that in my sample repo as well. The long names can also cause the Tekton variabele data store to explode

PR #185, opened 2024-11-27
change pipeline to tenantPipeline

tnevrlka:
- release-service now expects `tenantPipeline` instead of `pipeline`, see https://github.com/konflux-ci/release-service/blob/e091a3b077f7b1743c17869a043b8fd8856c6112/api/v1alpha1/releaseplan_types.go#L49
- using `pipeline` results in an `unknown field error` which leads to user confusion
- change `pipeline` to `tenantPipeline` to reflect the release-service change


yftacherzog (approved):
lgtm

arewm (approved):

PR #187, opened 2024-12-03
add activation keys docs

brianwcook:


docs/modules/ROOT/pages/how-tos/configuring/activation-keys-subscription.adoc
```diff
@@ -0,0 +1,53 @@
...
+4. Add a second key with the name *activationkey* and set the value to the name of your activation key.
+5. Save the secret by clicking *Add secret*
+
+[[Create-activation-key-through-console]]
+=== Create activation key secret through console
+
+Alternatively, you can create the secret through the CLI. After logging into your cluster and navigating to your namespace, run the following command:
+
+----
+kubectl create secret generic activation-key -n <your-tenant> --from-literal=org=<your org id> --from-literal=activationkey=<your activation key name>
```
arewm, line 35: This will make activation keys "just work" because of the default value of the parameter, correct? Would it be better if we suggested that an activation key is created with a unique name and require users to explicitly set the parameter (I guess they would need to set it on the pipelinerun, pipeline, and the task so that it is properly passed)?

docs/modules/ROOT/pages/how-tos/configuring/activation-keys-subscription.adoc
```diff
@@ -0,0 +1,53 @@
...
+----
+
+
+== Using subscription content in a build
+
+=== Automatic registration
+
+The buildah task will use a provided activation key to register itself with Red Hat subscription manager and moount the necessary credentials so that can be used by the builds.
```
arewm, line 43: If we use suggest the default activation key, can we inform users how to override the parameter so that a configured activation key isn't used everywhere?

arewm (approved):

docs/modules/ROOT/pages/how-tos/configuring/activation-keys-subscription.adoc
brianwcook: I added a new section to explain how naming impacts the scope of the key

docs/modules/ROOT/pages/how-tos/configuring/activation-keys-subscription.adoc
brianwcook: instead, I added a section advising the user to not use the default name unless they _want_ the key to be used for all builds in the workspace.  The pattern of adding a "dummy name" pointing to a non-existent key does work, but I think it seems like a weird thing to instruct people to do

PR #188, opened 2024-12-05
doc: custom timeout annotations for integration tests

dirgim:
* It is possible to set custom pipeline timeouts via annotations

Signed-off-by: dirgim <kpavic@redhat.com>


arewm (approved):
looks good to me.

Nit: should we include information on this in editing too? Or put this information there and cross ref it from creating? The action to take feels like more of an editing action since it is specified as a CLI-based annotation added to the ITS.

hongweiliu17:
> looks good to me.
> 
> Nit: should we include information on this in editing too? Or put this information there and cross ref it from creating? The action to take feels like more of an editing action since it is specified as a CLI-based annotation added to the ITS.

Look good to me. I have the same thought. Can we also permit users to edit integragtionTestScenario and set the annotation for the timeouts.

hongweiliu17 (approved):

dirgim:
@arewm @hongweiliu17 ack, good points, I updated the PR to have the instrunctions as part of the editing doc and cross-referenced it in the custom test creation.

jsztuka (approved):
lgtm

arewm (approved):

PR #190, opened 2024-12-06
added information about prefetching RPMs that require a subscription

brianwcook:


arewm (commented):
The PR generally feels good. Using the term "normal" feels odd. Is there some other way that we can say it? Can we just say non-hermetic?

brianwcook:
how about `non-hermetic (network-connected) builds:`

arewm:
Let's try that.

PR #193, opened 2024-12-11
add headings so page so that navigation is better

brianwcook:


arewm (approved):
Is this the only page that you saw needed the adjustments or it was just the one you were contending with?

PR #195, opened 2024-12-12
prefetching-dependencies update

slimreaper35:


docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -18,19 +18,22 @@ For every build, Cachi2 generates a software bill of materials (SBOM) where all
...
-|`rpm`
+|xref:bundler[Ruby]
+|`bundler`
 
-|xref:generic[Generic dependencies]
+|xref:rpm[N/A]
+|`rpm*`
+
+|xref:generic[N/A]
 |`generic fetcher`
```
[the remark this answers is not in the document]
slimreaper35: How does it look now?
(I swapped the columns in the table)

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -4,33 +4,39 @@ In {ProductName}, you can run a hermetic build by restricting network access to
...
+|`Ruby`
+
+|xref:rpm*[rpm]
+|`N/A`
+
+|xref:generic[generic]
+|`N/A`
 |===
 
 NOTE: To use Yarn as the package manager, see the link:https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#example-yarn[Cachi2 Usage guide].
```
arewm, line 37: Do we need this note still? If so, does it make sense to put it in the yarn section instead of up here?

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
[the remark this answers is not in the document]
arewm: That looks good, thanks!

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
slimreaper35: No, it should be dropped :disappointed: Thanks

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -298,6 +302,58 @@ spec:
 * From the {ProductName} *Applications* view, go to *Activity > Latest commits*.
 //include::../partials/con_hermetic_verification.adoc[]
 
+== [[yarn]]Enabling prefetch builds for `yarn`
+
+Supported versions: 1.x and 3.x. Cachi2 automatically detects the version of `yarn` and fetches any dependencies you declare in your `package.json` and `yarn.lock` project files.
+
+.Prerequisites
+
+* You have an up-to-date `yarn.lock` file in your source repository. To make sure that you have the latest `yarn.lock` file, or to create a lockfile, run the `yarn install` command. If `yarn.lock` is not up-to-date, Cachi2 will not fetch the dependencies.
```
arewm, line 311: Will Cachi2 not fetch any dependencies if the lockfiles are not up to date? Or is the behavior just that the locked dependencies will be fetched (and therefore your desired latest dependencies might not be fetched properly)?

(the comment is relevant for other package managers too)

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
```diff
@@ -298,6 +302,58 @@ spec:
...
+
++
+[source,yaml]
+----
+spec:
+  params:
+    - ...
+    - name: prefetch-input
+      value: '{"type": "yarn", "path": "."}'
+----
```
arewm, line 328: Is this basically the same across all package managers? Would it be possible to abstract the general form to a common location and then just have sections on the individuality of the different managers?

arewm (approved):
Thanks. This looks good overall. A couple of comments that would be nice to resolve but are non-blocking.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
slimreaper35: **TL;DR** It will fetch for `bundler` but not for `yarn`. I have to fix it...

It depends on the implementation details of each package manager.  For example for `yarn`, we use native mechanisms to pre-fetch the dependencies, the lockfile must be up to date. We use the command:  `yarn install --frozen-lockfile ...`  which fails in that case. For other package managers like `pip`, we just manually parse `requirements-txt` so it does not make sense to do any validation there.

docs/modules/ROOT/pages/how-tos/configuring/prefetching-dependencies.adoc
slimreaper35: Yes, there is *a lot* of repetition in the docs. Each package manager has the same paragraph:
```
Verification
From the Konflux Applications view, go to Activity > Pipeline runs.

Go to the pipeline run with Build in the Type column and confirm that the prefetch-dependencies stage displays a green checkmark. This indicates that the build process successfully fetched all dependencies.

From the Konflux Applications view, go to Activity > Latest commits.
```
So I would propose a follow-up to clean this up and overall make it more concise. Unless you favor doing it here.
But I can delete the `Procedure` section since it is duplicated for all package managers as well.

arewm: A follow up is fine if you don't want to do it righ tnow. You are being consistent.

PR #206, opened 2025-01-14
Change the name of the snyk secret to create

arewm:
The default secret name for the task is `snyk-secret`: https://github.com/konflux-ci/build-definitions/blob/9f209df8c12e8b321e63b77b5c91cb1523080a59/task/sast-snyk-check-oci-ta/0.3/sast-snyk-check-oci-ta.yaml#L49-L51

Therefore, in order to make it work out of the box, that should probably be the name that we suggest for users to configure.


dirgim (approved):
LGTM

PR #207, opened 2025-01-15
docs(RELEASE-1320): update tenant pipeline example pipelineRef

johnbieren:
The pipelineRef in the tenant pipeline never really made much sense. It suggested using a managed pipeline, which would not work (and many users made this mistake). This updates the example to something that should be more obviously an example and not a real pipeline.

We also moved the pipeline definitions in release-service-catalog, so the old reference won't even exist going forward. That is another reason to change it.

If a working tenant pipeline is created, we can update the docs to use that as the example.

### DIFF
--- a/docs/modules/ROOT/pages/advanced-how-tos/releasing/tenant-release-pipelines.adoc
+++ b/docs/modules/ROOT/pages/advanced-how-tos/releasing/tenant-release-pipelines.adoc
@@ -48,11 +48,11 @@ spec:
      resolver: git
      params:
        - name: url
-         value: "https://github.com/konflux-ci/release-service-catalog.git"
+         value: "https://github.com/<your-github-user>/<your-pipeline-repo>.git"
        - name: revision
-         value: production
+         value: main
        - name: pathInRepo
-         value: "pipelines/push-to-external-registry/push-to-external-registry.yaml"
+         value: "<path-to-your-pipeline>"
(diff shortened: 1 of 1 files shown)


johnbieren:
@arewm PTAL

docs/modules/ROOT/pages/advanced-how-tos/releasing/tenant-release-pipelines.adoc
```diff
@@ -48,11 +48,11 @@ spec:
      resolver: git
      params:
        - name: url
-         value: "https://github.com/konflux-ci/release-service-catalog.git"
+         value: "https://github.com/github-user/pipeline-repo.git"
```
arewm, line 51: Is this just a change to indicate that we don't have an out-of-the-box pipeline ready for tenant release examples?

If so, can we make this more explicit? Something like `https://github.com/<your-github-user>/<your-pipeline-repo>.git`

johnbieren: updated

johnbieren: Once https://github.com/konflux-ci/release-service/issues/589 is done, its result could go here. But for now, it doesn't make sense as written

arewm (approved):

PR #218, opened 2025-01-30
Add arch explanation to rpms.in.yaml

mmilev-rh:
While using the rpm-lockfile-prototype command when building a multi-arch container, all architectures must be specified using the `--arch` command-line option, otherwise the build task will fail.


ralphbean:
I think it may actually be preferable to specify the arches in `rpms.in.yaml` so that automated processes (like renovate) can know how to rebuild your rpms.lock.yaml file without having to somehow otherwise know which architectures to request.

Here's an example: https://github.com/konflux-ci/yq-container/commit/90c469600933d780ec1ff786111a101a9fc97fd8

ralphbean:
If you agree, can you update this patch that way?

mmilev-rh:
@ralphbean, patch has been updated as suggested.

ralphbean (approved):
Awesome, thank you!

PR #220, opened 2025-01-31
Document manual task update

lcarva:
Add a documentation block guiding users on how to update tekton bundle references manually without waiting for renovate.

### DIFF
--- a/docs/modules/ROOT/pages/troubleshooting/index.adoc
+++ b/docs/modules/ROOT/pages/troubleshooting/index.adoc
@@ -157,3 +157,82 @@ kubectl get secrets -o json |
         {auths: .}
     '
 ----
+
+== Manually Update Task Bundles
+
+Usually, Konflux users rely on link:https://docs.renovatebot.com/[renovate] to update
+the various Task bundle references in the build Pipelines. However, it is also possible
+to update these references manually if needed. For example, consider a build Pipeline
+that includes the following Task:
+
+[source,yaml]
+----
+- name: init
+  params:
+    - name: image-url
+      value: $(params.output-image)
+    - name: rebuild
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Nice :taco:

chmeliik (approved):

lcarva:
For posterity, this was ~stolen~ borrowed from https://github.com/enterprise-contract/hacks/blob/main/update-tekton-task-bundles.sh.

docs/modules/ROOT/pages/troubleshooting/index.adoc
```diff
@@ -157,3 +157,47 @@ kubectl get secrets -o json |
         {auths: .}
     '
 ----
+
+== Manually Update Task Bundles
+
+Usually, Konflux users rely on link:https://docs.renovatebot.com/[renovate] to update
```
arewm, line 163: @staticf0x , I think that you have some action item to move MintMaker documentation into this repo. We should update this comment/link during that process.

docs/modules/ROOT/pages/troubleshooting/index.adoc
```diff
@@ -157,3 +157,47 @@ kubectl get secrets -o json |
...
+the various Task bundle references in the build Pipelines. However, it is also possible
+to update these references manually if needed. The script below provides a working
```
arewm, line 165: Can you provide an overview of what this is doing? While what you are describing is manual, it is a self-triggered automated approach. Describing an _actual_ manual approach and then following up with this script might be helpful as well?

lcarva: Done.

arewm (approved):

PR #223, opened 2025-02-05
Fix vale issues

MartinBasti:


MartinBasti:
my IDE accidentally fixed trailing newlines, I can remove the change if needed.

PR #226, opened 2025-02-10
Enable task update automation to work on a mac

arewm:
Without the `-e` flag for sed, the command will fail on a mac with an `undefined label` error. Also changed the script name to be consistent throughout.


lcarva (approved):

dirgim (approved):
LGTM

PR #228, opened 2025-02-17
Minor fixes

tkdchen:
- Fix section level
- Fix annotation and URL in examples


tkdchen:
BTW,  there is an annotation `pipelinesascode.tekton.dev/sha-url`, use it in the example instead?

arewm:
@lcarva , what are your thoughts on whether we should use `sha-url` instead?

lcarva:
> @lcarva , what are your thoughts on whether we should use `sha-url` instead?

I can't seem to find it in the docs. What values does it hold?

tkdchen:
It holds the link of commit directly, e.g.
```
"pipelinesascode.tekton.dev/sha-url": "https://github.com/cqi-stonesoup-test/multi-arch-builds/commit/88d34991c95920b95dbcd6653eb42f5543bc5b87",
```

MartinBasti:
to use `sha-url` looks better to me

arewm (approved):

PR #236, opened 2025-02-26
KONFLUX-6912: Generate API Reference Docs

adambkaplan:
Initial attempt to generate reference documentation for Konflux APIs. This adds scripting to generate reference docs for Kubernetes CRDs that are added as part of a Konflux instance. The process utilizes the `crd-ref-docs` tool to generate asciidoc directly from source code.

Generation of equivalent reference docs for upstream dependencies, such as Tekton, is out of scope for this iteration. Future improvements may consider providing links to upstream sources.


ralphbean (approved):
This is great, thank you!

docs/modules/ROOT/pages/reference/_nav.adoc
```diff
@@ -0,0 +1,7 @@
+* xref:reference/index.adoc[Reference]
```
arewm, line 1: Do you want to move the glossary into the reference section as well?

arewm (commented):
How will we keep this up to date? Will someone just need to run it periodically?

Should we remove the API documentation from the architecture repo?

docs/modules/ROOT/pages/reference/kube-apis/application-api.adoc
```diff
@@ -0,0 +1,680 @@
...
+- xref:{anchor_prefix}-github-com-konflux-ci-application-api-api-v1alpha1-snapshotlist[$$SnapshotList$$]
+
+
+
+[id="{anchor_prefix}-github-com-konflux-ci-application-api-api-v1alpha1-application"]
+==== Application
+
+
+
+Application is the Schema for the applications API.  For a detailed description with examples, refer to <a href="https://github.com/redhat-appstudio/book/blob/main/book/HAS/hybrid-application-service-api.md"> Hybrid Application Service Kube API </a>
```
MartinBasti, line 33: This is not proper adoc link
image

docs/modules/ROOT/pages/reference/_nav.adoc
adambkaplan: I like it, but would consider this as a follow-up action. I'm working on a broader docs re-arch/reorg proposal.

docs/modules/ROOT/pages/reference/kube-apis/application-api.adoc
adambkaplan: Agreed. Since this is auto-generated content, I'd consider fixing this out of scope in this PR. I plan on issuing a "call to action" for teams that have their API docs auto-generated.

adambkaplan:
> How will we keep this up to date? Will someone just need to run it periodically?

Near term, someone would need to periodically re-generate the docs.

Note that the EC controller has a validation pattern that causes `crd-ref-docs` to render invalid asciidoc. I have a PR that fixes it [here](https://github.com/elastic/crd-ref-docs/pull/138), however I am not sure how active the project is and when a new release will be available.

> Should we remove the API documentation from the architecture repo?

Yes - the mechanisms to update that doc appears to be broken, plus some of those components may be deprecated/removed from Konflux.

PR #239, opened 2025-02-27
Add documentation about image annotations

Zokormazo:
Documents https://github.com/konflux-ci/build-definitions/pull/1980

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/_nav.adoc
+++ b/docs/modules/ROOT/pages/how-tos/_nav.adoc
@@ -11,7 +11,7 @@
 *** xref:how-tos/configuring/redundant-rebuilds.adoc[Preventing redundant rebuilds]
 *** xref:how-tos/configuring/build-with-args.adoc[Passing buildah arguments]
 *** xref:how-tos/configuring/custom-tags.adoc[Using custom tags]
-*** xref:how-tos/configuring/dynamic-labels.adoc[Using dynamic labels]
+*** xref:how-tos/configuring/labels-and-annotations.adoc[Using labels and annotations]
 *** xref:how-tos/configuring/overriding-compute-resources.adoc[Overriding compute resources]
 *** xref:how-tos/configuring/activation-keys-subscription.adoc[Using Red Hat subscription content]
 *** xref:how-tos/configuring/reconfiguring-build-pipeline.adoc[Reconfiguring the build pipeline]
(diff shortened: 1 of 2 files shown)


arewm (approved):
I love it, thanks!

docs/modules/ROOT/pages/how-tos/_nav.adoc
```diff
@@ -12,6 +12,7 @@
 *** xref:how-tos/configuring/build-with-args.adoc[Passing buildah arguments]
 *** xref:how-tos/configuring/custom-tags.adoc[Using custom tags]
 *** xref:how-tos/configuring/dynamic-labels.adoc[Using dynamic labels]
```
arewm, line 14: After thinking about this a little more, how would you feel about merging this new content with the labels page? That page isn't _entirely_ about dynamic labels. We could change the name of it to just refer to build metadata, defining labels and annotations, or something similar. After that, we would have a section on labels, annotations, and then dynamic labels (with further subsections).

The dynamic labels are just just key-value pairs, so we might even be able to take advantage of the dynamic labels flow to add dynamic annotations.

Zokormazo: That's a great idea, it should work with both

Zokormazo: Better now?

docs/modules/ROOT/pages/how-tos/configuring/labels-and-annotations.adoc
```diff
@@ -29,16 +29,47 @@ tasks:
...
+            value: generate-labels
+          - name: bundle
+            value: quay.io/konflux-ci/tekton-catalog/task-generate-labels:0.1@sha256:0068fe8b3c5d010daee5a922781a74cfb82251e775c260d14d9e50dd1a7aca65
+          - name: kind
+            value: task
+        resolver: bundles
+...
+----
 
+You can add the task twice to generate dynamic labels and annotations:
```
arewm, line 72: Can you not just use one invocation and reapply the same properties on both annotations and labels (this is what you describe in the `combining approaches` section at the end)?

Or are you saying that you can invoke the `generate-labels` task to create separate sets of dynamic annotations and labels?

docs/modules/ROOT/pages/how-tos/configuring/annotations.adoc
```diff
@@ -0,0 +1,30 @@
+= Using annotations
```
arewm, line 1: Did you mean to delete this file?

docs/modules/ROOT/pages/how-tos/configuring/labels-and-annotations.adoc
Zokormazo: oh, that's a great point. Yeah, my initial thought was that I would need to use the generate-labels twice to get labels and annotations, but you are right, you can create the full list once and then pass some values to one param and some others to the other. Let me update the doc for this :)

arewm (approved):
Thanks a lot for this documentation!

PR #240, opened 2025-02-27
fix: tenant pipelines should use own role instead of cluster role

johnbieren:


johnbieren:
@davidmogar @theflockers PTAL

arewm (approved):

johnbieren:
Thanks Andrew!

PR #241, opened 2025-02-27
Workflows: remove size labeler

MartinBasti:
This worklow doesn't work and I don't see much value in it. Instead of fixing it lets drop it.


MartinBasti:
If there is a wish to keep it, I can fix permissions

arewm (approved):
Thanks. I have left it because I wanted to fix it. We can always re-add it later.

PR #242, opened 2025-03-04
Change testing farm docs link

ukulekek:
Testing Farm Team needed to move the repo because the ArtifactsHub doesn't support subgroups in GitLab


ralphbean:
Thanks!

PR #243, opened 2025-03-06
Add a doc on inspecting scan results

ralphbean:


docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
```diff
@@ -0,0 +1,94 @@
...
+Note, other tools are gaining support for working with non-image OCI artifacts, including link:https://docs.podman.io/en/latest/markdown/podman-artifact.1.html[podman].
+
+== Discovering the scan results
+
+You can view all associated artifacts with `oras discover`:
+
+[source]
+--
+❯ oras discover $IMAGE
+quay.io/konflux-fedora/ralph-tenant/nethack@sha256:391c5670843d655d8f5d374aafe6367a6f450a10d2d95ebc123f65fe1479813f
```
arewm, line 19: Do you want to use an image from one of our sample or Konflux components instead?

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
```diff
@@ -0,0 +1,94 @@
...
+├── application/sarif+json
+│   ├── sha256:023d791628821f1fe429ce54495f2289b8ab59e990a8fd51e5499586b03d283a
+│   └── sha256:a9898962b1d7be572b026b1866e2422075afa19778ddd7119711c7f5dbbb7851
+└── application/vnd.clamav
+    └── sha256:0ad591d985039df3d3eaf13fd68026effc054c23d91467247fc35caf28fb12f3
+--
+
+And, you can format that as json so that you can work with it with tools like `jq` to do useful things.
```
arewm, line 27: Do you want to add more information about the "useful things" or leave that phrase out? Maybe just change it to enable programatic processing of the output?

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+❯ ls *.sarif
+sast_unicode_check_out.sarif  shellcheck-results.sarif
+--
+
+== Inspect the sarif files
+
+You can inspect those files locally, directly, or use them with other tools that support sarif.
+
+Here's a non-exhaustive list of some tools you might try to explore your sarif files:
```
arewm, line 88: I am not familiar with serif, but the serif examples above are pretty small. Why would users need custom tools to view these?

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
ralphbean: Sure!

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
ralphbean: The original user story that brought me to write this was a report that the logs of the sast scans were scary and overwhelming. I assume that for some software the sarif files might be enormous or unwieldy, but I don't have direct evidence of that.

kannon92: Hi, I'm the pesky reason for these docs..

I think this is really helpful as sarif logs are json files which can be difficult to read. 

Is there anyway to post process these files and have them as an artifact we could view on the scanner task?

ralphbean: View in the konflux UI? Yes, but it's a big feature not a quick fix. (In part because the the UI doesn't know how to speak OCI directly so there's some translation to do before we even get to rendering the results nicely: https://issues.redhat.com/browse/KONFLUX-4923)

Another route could be for us to post the sarif results back to github's codeql api, which would have them show up nicely in your PR.

kannon92: I guess I am prow guy. I was thinking we could have the artifacts that these generate attached somewhere that we could at least download and view.

Post progress them with oras, upload them to a directory and then have someone be able to at least view the raw logs.

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
ralphbean: Actually, I'm having trouble finding another good one. I'm going to stick with this one.

docs/modules/ROOT/pages/how-tos/metadata/scan-results.adoc
```diff
@@ -0,0 +1,94 @@
...
```
arewm, line 88:
```suggestion
You can inspect those files locally, directly, or use them with other tools that support sarif. Some examples of tools that might help to inspect large files include:
```

arewm (approved):
Added a suggested change. Ultimately it is a nitpick. We can iterate later if needed.

PR #256, opened 2025-03-14
List and link to the individual APIs by name

ralphbean:
This will hopefully be more consistent and recognizable to users.


MartinBasti (commented):
:rocket:

PR #266, opened 2025-03-17
Move MintMaker documentation upstream

staticf0x:
This is the whole documentation on dependency management / MintMaker from internal docs. I removed any reference to internal stuff, like Slack channels or Jira issues, that would not be accessible from the public web.


arewm (approved):
This looks great, thanks.

Are there any clear places where you might want to extend/modify this documentation to add additional content (i.e. when rebuilding it internally)? If so, we can use partials for that content.

PR #269, opened 2025-03-17
Fix other codeowners references

ralphbean:
This is a follow up to https://github.com/konflux-ci/docs/pull/268 after we confirmed that it worked as expected.


dirgim (approved):
LGTM

PR #275, opened 2025-03-18
feat(STONEINTG-1129): create user guide for monorepo users

dirgim:
* Add additional guidance for testing and releasing the monorepo pattern applications

Signed-off-by: dirgim <kpavic@redhat.com>

### DIFF
--- a/modules/patterns/pages/managing-monorepo-applications.adoc
+++ b/modules/patterns/pages/managing-monorepo-applications.adoc
@@ -59,7 +59,7 @@ spec:
             fieldRef:
               fieldPath: metadata.annotations['pac.test.appstudio.openshift.io/sha']
       script: |
-        #!/bin/sh
+        #!/bin/bash
         set -e
         components=$(jq -c '.components[]' <<< "${SNAPSHOT}")
         snapshotUrlUrlWithoutSuffix=$(echo $SNAPSHOT_URL | sed 's/\.git$//')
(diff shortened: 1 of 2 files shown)


jencull (changes requested):
Nit: remove 'ok' in this line: 

`If the change to components does not span across multiple components, the above workflow will work ok in most cases.`

jencull (approved):
lgtm

kasemAlem (approved):

modules/patterns/pages/managing-monorepo-applications.adoc
```diff
@@ -0,0 +1,81 @@
...
+        - name: SNAPSHOT_URL
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['pac.test.appstudio.openshift.io/source-repo-url']
+        - name: SNAPSHOT_SHA
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['pac.test.appstudio.openshift.io/sha']
+      script: |
+        #!/bin/sh
```
kasemAlem, line 62: `[[ ]]` is not supported in #!/bin/sh and we should use `#!/bin/bash` instead,  wdyt?

sonam1412 (approved):
lgtm

ralphbean:
This LGTM, but I think a great addition would be something related to using `on-cel-expression` to limit the build.

Let's merge this as is, but then follow up to add a link out to `modules/building/pages/redundant-rebuilds.adoc`

modules/patterns/pages/managing-monorepo-applications.adoc
ralphbean:
```suggestion
        #!/bin/bash
```

ralphbean (approved):

arewm (approved):
Not sure why this hasn't auto-merged yet. I agree with what @ralphbean said though.

PR #276, opened 2025-03-19
doc: schedule is specific to package rule

yftacherzog:
Renovate schedule can be set in package rules rather than manager configs.


staticf0x:
Both is possible, if you take a look at the default config: https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json we set a different schedule for different managers, but it's also possible to set on `packageRules` level, however that's not set anywhere in the default config.

yftacherzog:
> Both is possible, if you take a look at the default config: https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json we set a different schedule for different managers, but it's also possible to set on `packageRules` level, however that's not set anywhere in the default config.

Thanks. Good to know!

PR #277, opened 2025-03-19
Document how to create activation key secrets to be usable by MintMaker

querti:
MintMaker cannot match custom activation key secrets to their repositories without extra labels and annotations. Document how users can create such secrets.

The labels and annotations are analogous to Gitlab secret annotations[1].

[1] https://konflux-ci.dev/docs/building/creating-secrets/#gitlab-source-secret

Refers to CLOUDDST-26090


staticf0x (approved):
Sounds good and understandable :+1:

staticf0x (approved):

modules/building/pages/activation-keys-subscription.adoc
```diff
@@ -45,6 +52,82 @@ Alternatively, you can create the secret through the CLI. After logging into you
 kubectl create secret generic activation-key -n <your-tenant> --from-literal=org=<your org id> --from-literal=activationkey=<your activation key name>
 ----
 
+[[Create-custom-activation-key-secret]]
```
arewm, line 55: Would it make sense to just combine with with the section above, `=== Create activation key secret through console`?

arewm: Sounds good. I guess the other solution is much simpler in that you don't need to interact with creating CRs directly.

PR #278, opened 2025-03-19
Fix title and callouts on labels and annotations page

Zokormazo:
I realized today while checking the new structure I made two mistakes while extending the labels docs with annotations.

First of all, I forgot to change the page title. I changed the section name on the ToC, but not the name on the table itself.
The second one is about the callouts on the code. The callout on the code block are numbered `1` and `2` for labels and annotations but then there is a single explanation for both of them


arewm (approved):
Thanks for the fixes!

PR #280, opened 2025-03-19
fix: add mention of permissions needed for retest

scoheb:


arewm (commented):
Link highlighting like this is very fragile. If the page is modified to edit the text selected then the highlighting might break (not sure if it is cross-browser either).

What would you say to just linking to the page itself noting that the `/retest` permissions follows the same requirements as determining whether a submitter is allowed to run the PipelineRun?

arewm (approved):

PR #285, opened 2025-03-21
prefetching-dependencies: Add Cargo (Rust) to supported package managers

slimreaper35:


tnevrlka (approved):

arewm (approved):

tnevrlka:
My bad, it seems like the Cargo support is not released yet. Do we want to roll this back, could it confuse users who would now be expecting it to work?

PR #291, opened 2025-03-26
Add `finally` nuance to slack-webhook-notification

schmidtd:
Simply adding the slack-webhook-notification task as suggested in the original doc will fail with a message like this:

```Tekton Controller has reported this error: admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: pipeline tasks can not refer to execution status of any other pipeline task or aggregate status of tasks: spec.pipelineSpec.tasks[2].when[0]```

A pipeline that fans out needs to rendezvous and aggregate the status *at the end* to be sure it can catch any failure along the line.


arewm (approved):
Thanks!

PR #292, opened 2025-03-28
Rename managing-compliance-with-ec to just compliance

ralphbean:
If you look at the urls, all of the modules have neat tidy short names except for this one.

Here, make it shorter to match the style of the others and change some of the sub-page titles to also match the shorter style.


ralphbean:
Downstream, I want to move our policy-exceptions doc to a nav partial here, but I'll do that in a followup PR.

ralphbean:
See also https://gitlab.cee.redhat.com/konflux/docs/users/-/merge_requests/308

arewm (approved):
LGTM but it will be another breaking change. Do we have any references to this URL in the UI or build-definitions? I had to fix some in both of those after our previous changes.

ralphbean:
At least downstream, we can add `_redirects` for them. Let me grep around first.

ralphbean:
(I haven't found any hits so far.)

PR #298, opened 2025-04-01
Document that component name has to be unique in a namespace

rcerven:
STONEBLD-3083


modules/ROOT/pages/getting-started.adoc
```diff
@@ -35,6 +35,7 @@ Snapshot which will trigger a specific pipeline in the managed namespace.
 A Component CR describes the properties for an OCI artifact including the git repository where the artifact is from,
 its latest built commit, initial build configuration parameters (i.e. Containerfile path), and any relationships to
 other Components. The CR also contains a reference to its single owning Application.
+Component name has to be unique in a namespace, even though it would be in different application.
```
arewm, line 38: How about making a change like this to both places?
```suggestion
Component names must be unique in a namespace even components are in separate applications.
```

mmorhun (approved):

modules/ROOT/pages/getting-started.adoc
rcerven: changed

arewm (approved):

PR #299, opened 2025-04-02
Reorganize troubleshooting to include partials

ralphbean:
I organized the troubleshooting doc into three sections: build, release, and registries and made partials available for each section to allow downstream customization.

### DIFF
--- a/modules/building/pages/creating-secrets.adoc
+++ b/modules/building/pages/creating-secrets.adoc
@@ -103,7 +103,7 @@ type: kubernetes.io/dockerconfigjson
     - RBAC: The ServiceAccount must have `get` permission for the secret.
 
 * **Troubleshooting**
-  For issues with secret linking, review the xref:troubleshooting:index.adoc#check-if-the-secret-is-linked-to-the-service-account[troubleshooting section].
+  For issues with secret linking, review the xref:troubleshooting:registries.adoc#check-if-the-secret-is-linked-to-the-service-account[troubleshooting section].
 ====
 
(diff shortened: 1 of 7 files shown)


ralphbean:
See downstream change with relevant partials at https://gitlab.cee.redhat.com/konflux/docs/users/-/merge_requests/311

arewm (commented):
Do you want to break these into their own pages to prevent the troubleshooting from turning into a single large page?

ralphbean:
> Do you want to break these into their own pages to prevent the troubleshooting from turning into a single large page?

I don't know. We have partials now, so I suppose I'm playing with using them to see how things work. How much is too much?

If I look at the kube docs, they have a single section for troubleshooting, but it is broken out into many pages.

If I look at the tekton docs, they have a troubleshooting or debug page nested in underneath each of their other sections (one for triggers, one for pipelineruns/taskruns, etc.)

ralphbean:
Thinking about what a more-developed troubleshooting section would look like for us: it would at least have a section for troubleshooting an _installation of konflux_ and another section for troubleshooting builds and releases as a user. Maybe it does make sense to split this now into separate pages rather than a single page with partials.

ralphbean:
Ok, that's better.

arewm (approved):
Thanks. I like this approach!

We had discussions a while ago about locating troubleshooting in each section or pulling it all together. Some of our sections are really long (i.e. building) so I feel like troubleshooting will get lost there easily.

MartinBasti (approved):

PR #309, opened 2025-04-15
Enable Konflux bot to autorun prow tests

fmudry:
Users wanted to know how to autorun prow tests when Konflux creates sync pull requests.
This was the solution proposed in a konflux support thread.


fmudry:
You are right, after reconsideration it'll be better to move this to the downstream documentation.

MartinBasti:
prow is public k8s thing, why to move it only to downstream?

PR #313, opened 2025-04-17
Add Note to configure-pac in ProjectDevStreamTemplate

MartinBasti:


modules/patterns/pages/managing-multiple-versions.adoc
```diff
@@ -43,6 +43,11 @@ spec:
 
 To enable quickly creating multiple development streams, we must create a template for them.
 
+[NOTE]
```
arewm, line 46: Do you want to make this more apparent like a warning?

MartinBasti: isn't it too strong? I'd keep it for more breaking changes

arewm: How about CAUTION (I think that is one too)?

yftacherzog (dismissed):

modules/patterns/pages/managing-multiple-versions.adoc
```diff
@@ -74,7 +79,6 @@ spec:
     metadata:
       annotations:
         build.appstudio.openshift.io/pipeline: '{"name":"docker-build","bundle":"latest"}'
-        build.appstudio.openshift.io/status: '{"pac":{"state":"enabled","merge-url":"https://github.com/konflux-ci/multi-version-konflux-sample/pull/1","configuration-time":"Wed, 07 Aug 2024 08:59:18 UTC"},"message":"done"}'
```
ifireball, line 77: This needs to be there so the UI thinks PAC was already configured.

modules/patterns/pages/managing-multiple-versions.adoc
```diff
@@ -43,6 +43,11 @@ spec:
 
 To enable quickly creating multiple development streams, we must create a template for them.
 
+[NOTE]
+====
+Make sure that your component definition template doesn't have annotation `build.appstudio.openshift.io/request: configure-pac` configured. This will lead into repeated onboarding and suggesting new changes to your pipeline runs. Project Controller will do onboarding for you.
```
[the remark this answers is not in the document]
MartinBasti: but `configure-pac` should be omitted, or shouldn't?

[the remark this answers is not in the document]
MartinBasti: @ifireball ping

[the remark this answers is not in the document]
ifireball: it should be omitted, but we should ensure users understand not to expect pipeline PRs to be created.

modules/patterns/pages/managing-multiple-versions.adoc
ifireball: I think we already have some test about this elsewhere in the document

PR #317, opened 2025-04-29
Improve the documentation on how to use the RPM lockfiles with entitled repositories

brunoapimentel:


modules/building/pages/activation-keys-subscription.adoc
```diff
@@ -340,7 +340,7 @@ resolve to the correct path on the file system. These options point to
 certificates and keys that use a unique identifier (e.g., `+sslclientcert =
 /etc/pki/entitlement-host/$ID.pem+`). You may see SSL issues if you copied a
 repository configuration file from a different system registered with a
-different entitlement or activation key.
+different entitlement or activation key. Alternatively, these options can be set by using the `DNF_VAR_SSL_CLIENT_KEY` and `DNF_VAR_SSL_CLIENT_CERT` environment variables.
```
msugakov, line 343: Well, this _assumes_ the repo file already has been modified to use `sslclientkey = $SSL_CLIENT_KEY` and `sslclientcert = $SSL_CLIENT_CERT` but such change is not even mentioned on **this** page `modules/building/pages/activation-keys-subscription.adoc`.

I think, it's best to leave this out here and only modify `modules/mintmaker/pages/rpm-lockfile.adoc`.

modules/mintmaker/pages/rpm-lockfile.adoc
```diff
@@ -36,7 +36,7 @@ The RPM lockfile is also required to be located in the root folder of the reposi
...
+MintMaker is able to update RPM lockfiles containing RPMs that require subscription, but some setup needs to be completed first. Firstly, activation key secret needs to be added to the user namespace. This process is described in detail in the xref:ROOT:building:activation-keys-subscription.adoc#configuring-an-rpm-lockfile-for-hermetic-builds[Using Red Hat activation keys to access subscription content] article. Secondly, the RPM repofile needs to be modified to make use of the custom authentication certificate and key. For each repository requiring authentication, please set `sslclientkey = $SSL_CLIENT_KEY` and `sslclientcert = $SSL_CLIENT_CERT`. This will ensure that MintMaker sets the correct paths for the authentication. For example:
 
 [source]
```
msugakov, line 41: I'd suggest to add a note after this source fragment saying something like the following

> If you still need to run `rpm-lockfile-prototype` manually against the repofile which was modified this way, you should set environment variables `DNF_VAR_SSL_CLIENT_KEY` and `DNF_VAR_SSL_CLIENT_CERT` to point to actual key and cert files before running `rpm-lockfile-prototype` command.

brunoapimentel: One thing that is not clear to me is if this issue only happened because you opted for a local setup (IOW, if you run the same container you're using for your build and subscribe it, I assume the key and cert would be in the right places, right?).

The reason I'm asking is that I'm trying to figure out if we need to clarify this is only needed for local setups (as the linked doc teaches how to do it based on a container approach).

staticf0x:
We already have https://github.com/konflux-ci/docs/pull/316

modules/mintmaker/pages/rpm-lockfile.adoc
msugakov: I'm running `rpm-lockfile-prototype` locally, in a container :-)
I described what I run in this message https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1745861163909809?thread_ts=1745849700.315909&cid=C04PZ7H0VA8

Start plain UBI9 container (for reproducibility) with repo source code mounted, activate subscription-manager (to get cert and key in place, this gives different file names each time), install `rpm-lockfile-prototype` there and run it against the files in the repo.

brunoapimentel: Oh, thanks for clarifying. I see it now.

PR #321, opened 2025-05-02
building binaries with Konflux docs

ernesgonzalez33:
Adding some documentation about building binaries with Konflux

### DIFF
--- a/modules/building/nav.adoc
+++ b/modules/building/nav.adoc
@@ -1,6 +1,7 @@
 ** xref:index.adoc[Building]
 *** xref:creating.adoc[Creating applications and components]
 *** xref:deleting.adoc[Deleting applications and components]
+*** xref:build-binaries.adoc[Building binaries]
 *** xref:imagerepository.adoc[Customizing ImageRepository]
 *** xref:customizing-the-build.adoc[Customizing the build pipeline]
 *** xref:creating-secrets.adoc[Creating secrets for your builds]
--- a/modules/building/pages/build-binaries.adoc
+++ b/modules/building/pages/build-binaries.adoc
@@ -1,0 +1,31 @@
+= Building Binaries in {ProductName}
+
+This guide provides guidelines on how to build binaries within {ProductName}.
+
+== Theory
+
+The fundamental process involves creating a Dockerfile to encapsulate your binary build process. This is similar to your existing build process, but it's executed inside a Docker container. You'll then define your application and component in {ProductName} as you normally would. For details on this, refer to xref:creating.adoc[Creating applications and components].
+
+If you're unfamiliar with Dockerfiles, consult the official link:https://docs.docker.com/reference/dockerfile/[documentation].
+
+== Example
+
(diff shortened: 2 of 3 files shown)


ernesgonzalez33:
I want to also create an internal version of this, so putting this as draft until that is done

ibuziuk (commented):
I believe the PR is missing a few critical SOPs: 

- The recommended way of building / organizing multi-arch build for darwin / linux / windows
- Process for uploading binaries to developers.redhat.com

ernesgonzalez33:
> I believe the PR is missing a few critical SOPs:
> 
>     * The recommended way of building / organizing multi-arch build for darwin / linux / windows
> 
>     * Process for uploading binaries to developers.redhat.com

Yes, because all of that is going to be in the internal one

ralphbean:
> I want to also create an internal version of this, so putting this as draft until that is done

This is awesome, thanks!  The best way to integrate with internal docs is to have only one main upstream doc, but with enough "partials" that you can fill out all extra downstream details that you need in partials there.

modules/building/pages/konflux-build-binaries.adoc
```diff
@@ -0,0 +1,19 @@
+= Building Binaries in {ProductName}
+
+This guide provides guidelines on how to build binaries within {ProductName}.
+
+== Theory
+
+The fundamental process involves creating a Dockerfile to encapsulate your binary build process. This is similar to your existing build process, but it's executed inside a Docker container. You'll then define your application and component in {ProductName} as you normally would. For details on this, refer to xref:creating.adoc[Creating applications and components].
```
arewm, line 7:
```suggestion
The fundamental process involves creating a Containerfile to encapsulate your binary build process, similar to the production of container images. This allows you to define applications and components in {ProductName} as you xref:creating.adoc[normally would].
```

modules/building/pages/konflux-build-binaries.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+If you're unfamiliar with Dockerfiles, consult the official link:https://docs.docker.com/reference/dockerfile/[documentation].
```
arewm, line 9: We should use `Containerfiles` but there isn't as clear of an authoritative documentation source as there is for a Dockerfile.

https://github.com/containers/buildah/blob/main/docs/buildah-build.1.md ?

modules/building/pages/konflux-build-binaries.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+== Example
+
+Let's assume you have a shell script named `build-binary.sh` in your repository that compiles your binary. A basic Dockerfile would resemble this:
```
arewm, line 13: Should you put this header within the partial?

modules/building/pages/konflux-build-binaries.adoc
ernesgonzalez33: What do you suggest then? I can change it to Container file, but I still prefer the Dockerfile reference docs. Maybe I can remove the link altogether

ernesgonzalez33: Fixed

arewm (approved):

PR #324, opened 2025-05-07
fix: Document Snapshot introductory material

arewm:
We just jump into working with Snapshots but we don't provide enough context about Snapshots and what they mean. We can improve that with a summary index.adoc.

Resolves: STONEINTG-1204

### DIFF
--- a/modules/testing/pages/integration/snapshots/index.adoc
+++ b/modules/testing/pages/integration/snapshots/index.adoc
@@ -1,0 +1,94 @@
+= Snapshots
+
+// Optional: Define context for cross-references if needed within the module structure
+// :context: snapshots
+
+A {ProductName} **Snapshot** represents a specific, immutable collection of container images for the Components within an Application at a particular point in time. They capture the exact versions of all relevant component artifacts that were built and intended to be used together including their Git commits.
+
+Snapshots are fundamental for ensuring consistency during testing and release processes. They provide a stable target for:
+
+* Running integration tests defined by IntegrationTestScenarios (ITS).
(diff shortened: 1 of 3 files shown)


yftacherzog (approved):

modules/testing/pages/integration/snapshots/index.adoc
```diff
@@ -0,0 +1,94 @@
+= Snapshots
+
+// Optional: Define context for cross-references if needed within the module structure
+// :context: snapshots
+
+A {ProductName} **Snapshot** represents a specific, immutable collection of container images for the Components within an Application at a particular point in time. They capture the exact versions of all relevant component artifacts that were built and intended to be used together including their Git commits.
```
dirgim, line 6: Should we move the information from the [What is a Snapshot section](https://github.com/konflux-ci/docs/blob/4508c8bcd473d004426fa97b0f3c4fbe1d6468bd/modules/testing/pages/integration/snapshots/working-with-snapshots.adoc?plain=1#L5-L31) here?

jsztuka (approved):
lgtm

modules/testing/pages/integration/snapshots/index.adoc
arewm: Yes, good call! I just pushed this with a new commit.

arewm: @dirgim , how does it look now?

dirgim: Looks good, thanks

dirgim (approved):
LGTM

PR #327, opened 2025-05-08
Change the website title to "Konflux Documentation"

brunoapimentel:
This change will affect the name that is displayed in the browser tab.

Since the added "Konflux" will be redundant with the logo that is displayed in the site's header (making it show as "Konflux Konflux Documentation"), we need to disassociate the rendering of the header with the site's configured title.


dirgim (approved):
LGTM

arewm (approved):

PR #332, opened 2025-05-13
Add vale checks for voice consistency

arewm:
While these checks have not been run against the repository as a whole, they will run automatically on asciidoc files that are modified in PRs.

As the files continue to be updated, we will slowly improve the voice consistency of the documentation.


ralphbean:
Super cool!

hack/vale-check.sh
```diff
@@ -0,0 +1,113 @@
...
+
+    # Run vale sync first
+    sync_vale
+
+    # Run vale with all arguments passed to the script
+    run_vale "$@"
+}
+
+# Pass all arguments to main
+main "$@"
```
MartinBasti, line 113: would be nice to keep posix endlines at the end of each file

.github/workflows/linters.yaml
```diff
@@ -13,14 +13,26 @@ jobs:
...
+        uses: tj-actions/changed-files@v41
+        with:
+          files: |
+            **/*.adoc
+      
       - name: Vale Linter
         uses: errata-ai/vale-action@reviewdog
+        if: steps.changed-files.outputs.any_changed == 'true'
         with:
-          vale_flags: "--minAlertLevel=error"
```
MartinBasti, line 21: Why this change?

MartinBasti: That's why
```
Konflux.ActiveVoice = warning
```

MartinBasti (approved):

.vale/styles/Konflux/ActiveVoice.yml
```diff
@@ -0,0 +1,14 @@
...
+tokens:
+  - is being
+  - are being
+  - was being
+  - were being
+  - has been
+  - have been
+  - had been
+  - will be
+  - being
```
MartinBasti, line 14: endlines are missing also in these yml files

.github/workflows/linters.yaml
```diff
@@ -13,14 +13,26 @@ jobs:
...
+          files: ${{ steps.changed-files.outputs.all_changed_files }}
+          vale_flags: "--no-exit --output=line"
```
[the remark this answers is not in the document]
arewm: Possibly... I was trying to use: https://github.com/reviewdog/reviewdog?tab=readme-ov-file#reporter-github-pullrequest-review-comment--reportergithub-pr-review

I need to look into this more.

.vale/styles/Konflux/DirectAddress.yml
```diff
@@ -0,0 +1,9 @@
+extends: substitution
+message: "Use 'you' instead of '%s' for direct address"
```
MartinBasti, line 2: This error message looks sus
```
modules/troubleshooting/pages/builds.adoc:40:88:Konflux.DirectAddress:Use 'you' instead of 'you' for direct address
```

MartinBasti, line 2: probably this should be the fix
```suggestion
message: "Use '%s' instead of '%s' for direct address"
```

package.json
```diff
@@ -11,10 +11,13 @@
   "scripts": {
     "api-gen": "./hack/gen-api-docs.sh",
     "build": "antora generate --clean --fetch antora-playbook.yml",
-    "preview:netlify": "netlify deploy --dir public"
+    "preview:netlify": "netlify deploy --dir public",
+    "lint:docs": "./hack/vale-check.sh",
+    "lint:docs:fix": "./hack/vale-check.sh",
+    "lint:docs:strict": "./hack/vale-check.sh --minAlertLevel=error",
```
MartinBasti, line 17: This sounds like opposite of strict check, at least I'm used to `strict=warnings are errors`

package.json
```diff
@@ -11,10 +11,13 @@
   "scripts": {
     "api-gen": "./hack/gen-api-docs.sh",
     "build": "antora generate --clean --fetch antora-playbook.yml",
-    "preview:netlify": "netlify deploy --dir public"
+    "preview:netlify": "netlify deploy --dir public",
+    "lint:docs": "./hack/vale-check.sh",
+    "lint:docs:fix": "./hack/vale-check.sh",
```
MartinBasti, line 16: what's the difference between there two? Why we have fix?

CONTRIBUTING.md
```diff
@@ -115,3 +115,71 @@ npm run api-gen
...
+
+The CI pipeline will fail on error-level violations but only warn about warning-level violations. This balance ensures critical voice consistency while maintaining flexibility where needed.
+
+To check your documentation against these rules:
+```bash
+# Show all issues (warnings and errors)
+npm run lint:docs
+
+# Check all documentation
+npm run lint:docs
```
MartinBasti, line 155: twice the same example?

MartinBasti (changes requested):

.vale.ini
```diff
@@ -14,6 +14,12 @@ Vocab = konflux
 
 BasedOnStyles = RedHat, Konflux
 
+# Voice consistency rules with recommended levels
```
MartinBasti, line 17: Those level are configured in the rules themselves, why do we override configuration here, instead of changin level in rules?

CONTRIBUTING.md
arewm: Thanks for catching this. It was a copy/paste issue. I will resolve it on the next push.

package.json
arewm: No difference. The fix target was added accidentally.

package.json
arewm: I think that there is also a benefit to only showing the errors as that is what will block PRs. Is there a better name that we can have to this script target?

arewm: I changed this to `lint:docs:errors`

CONTRIBUTING.md
```diff
@@ -115,3 +115,72 @@ npm run api-gen
...
+npm run lint:docs:errors
+npm run lint:docs:changed:errors
+```
+
+The script will:
+1. Check if all required dependencies are installed
+2. Run `vale sync` to ensure all style rules are up to date
+3. Run Vale checks with the same settings as the CI environment
+
+> **Note:** Pull requests will fail if any voice consistency errors are found. Use `npm run lint:docs:strict` to run the same checks locally that will be run in CI.
```
MartinBasti, line 174: `lint:docs:strict` this changed, and also it doesn't have effect to which check are running, only to output. And you removed `--minAlertLevel=error` from GH workflow so it doesn't match CI now even in output.

MartinBasti (approved):

ralphbean (approved):

PR #340, opened 2025-05-16
feat(KONFLUX-277) move slack notifications to its own page

ralphbean:
The purpose is to highlight it in search results, since people ask about it so much.


arewm (changes requested):
Seems reasonable as it is hidden in its current location. You should add this to the sidebar TOC though.

ralphbean:
Good catch! Done.

arewm (approved):

Issue #273, opened 2025-03-18
Improve getting-started.adoc


arewm:
Comment originally from @brunoapimentel 

> It feels weird that this section does not start by explaining what a namespace is. I'd add a very basic initial sentence just to define the concept of a Kubernetes namespace, and add a link to the official docs in case the person needs more context.

ref: https://github.com/konflux-ci/docs/pull/270/files#r1999665081

We should address this in a future change.

PR #341, opened 2025-05-16
Update the getting started doc

ralphbean:
In https://github.com/konflux-ci/docs/issues/273 we noted that the getting started doc was kind of bumpy.

I tried to smooth it out here.

Notably, I added an explanation of Custom Resources early on that I think was missing and I removed a good bit of text that I thought was unnecessary and cumbersome for the introduction doc.

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
 == Key concepts
 
 === Namespace
-A namespace in a Kubernetes cluster can have sole or shared ownership. As a user, it is the most global
-resource that is generally available to you. Most of the Kubernetes Custom Resources (CRs) which you interact
-with are scoped to a single namespace including Components, Applications, Snapshots, and Secrets.
-
-All builds are performed by Tekton PipelineRuns within a namespace and usually driven by PipelineDefinitions
-committed to git repositories and triggered via https://pipelinesascode.com/[Pipelines as Code].
+In Kubernetes,
+link:https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/[namespaces]
+provide a foundational mechanism for isolating groups of resources within a
+single cluster. All of the {ProductName} resources and APIs that you interact
+with are scoped to namespaces, including your Components, Applications, Snapshots,
+Secrets and the Tekton PipelineRuns that perform builds, tests, and releases.
 
 ==== Tenant namespace
 These are namespaces where artifacts are produced from Tekton Pipelines. These can be accessed by more than one
-individual according to users' roles and the permissions defined by 
+individual according to their roles and the permissions defined by 
 link:https://konflux-ci.dev/architecture/ADR/0011-roles-and-permissions.html[those roles]. The tenant namespaces can be either
 for an individual or a team.
 //TODO: Document the process for getting access to/creating new namespaces
 
-You can grant, modify, or remove other users access to any namespace for which you have the `admin` role. Each
-namespace's admins have the responsibility to ensure that appropriate access is granted to users. If users are
-over-privileged, they may have access to any added secrets or to perform other unintended actions.
+In {ProductName}, you operate in a tenant namespace which is scoped to your
+team. One team can span multiple namespaces if needed, each with many
+applications with components. Namespaces should NOT be shared by many teams.
+
+In your tenant namespace, you will create Components and Applications, run the
+build PipelineRuns that are defined in your git repositories, view and iterate
+on the results of your IntegrationTestScenarios, and create Releases to release
+specific Snapshots according to specific ReleasePlans.
+
+NOTE: You can grant, modify, or remove other users' access to any namespace for
+which you have the `admin` role. Each namespace's admins have the
+responsibility to ensure that appropriate access is granted to users. If users
+are over-privileged, they might have access to any added secrets or to perform
+other unintended actions.
 
 ==== Managed namespace
-Sometimes users will need to perform actions that require credentials in someone else's control. Since admins
-in a tenant namespace can see all secrets, these actions need to be performed via separate managed namespaces.
-The primary interaction mode between tenant and managed namespaces is to create a Release referencing a specific
-Snapshot which will trigger a specific pipeline in the managed namespace.
+These are namespaces where a privileged release engineering team or
+site-reliability engineering team manages release pipelines and credentials for
+your organization.
+
+The primary interaction mode between tenant and managed namespaces is to create
+a Release in a tenant namespace referencing a specific Snapshot which will
+trigger a specific release pipeline in the managed namespace.
+
+=== Custom Resource (CR)
+In Kubernetes, a
+link:https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/[Custom
+Resources (CR)] is an extension of the Kubernetes API.
+
+All {ProductName} APIs are implemented as Kubernetes CRs. If
+you're familiar with common Kubernetes resources such as Pods and Deployments,
+you'll find that {ProductName} resources appear on cluster in the same way:
+Applications, Components, Snapshots, and PipelineRuns.
+
+A consequence of this fact is that commonplace Kubernetes client tools such as
+link:https://kubernetes.io/docs/reference/kubectl/[kubectl] understand and can
+work with {ProductName} APIs.
 
 === Component
 A Component CR describes the properties for an OCI artifact including the git repository where the artifact is from,
-its latest built commit, initial build configuration parameters (i.e. Containerfile path), and any relationships to
+its latest built commit, initial build configuration parameters, and any relationships to
 other Components. The CR also contains a reference to its single owning Application.
-Component names must be unique in a namespace, even when components are used in different application.
-
-=== Pipelines as Code (PAC)
-https://pipelinesascode.com/[Pipelines as Code] is a project that provides opinionated tooling to define a
-Tekton-based CI/CD pipeline. It enables the use of a Tekton PipelineDefinition within a git repository and for
-that pipeline to be triggered upon new commits/PR to the repository. In Konflux, these pipelines are pushed to
-the git repositories in the `.tekton` directory.
+Component names are unique in a namespace, even when components are used in different application.
 
 === Build pipeline
-Upon a new push or pull request (merge request in GitLab) event, the pipeline defined in the repository will be
+When you create a Component in {ProductName}, a build pipeline is pushed to the git repository in the `.tekton/`
+directory and a webhook is installed.
+
+Upon a new push or pull request (merge request in GitLab) event, the pipeline defined in the repository is
 run. This pipeline describes the process to build and test a specific artifact. The build process includes Tekton
-Tasks like cloning the git repository, prefetching dependencies, building the OCI artifact and source SBOM, and
-generating the source container. The test process includes Tekton Tasks like running Snyk scans, checking for
+Tasks such as cloning the git repository, prefetching dependencies, building the OCI artifact and source SBOM, and
+generating the source container. The test process includes Tekton Tasks such as running Snyk scans, checking for
 CVEs with clair-in-ci, and running an antivirus scan on the artifact.
 
+{ProductName} inherits the pattern of defining Tekton build pipelines in git
+from its use of https://pipelinesascode.com/[Pipelines as Code (PaC)]. PaC
+enables the use of a Tekton PipelineDefinition within a git repository and for
+that pipeline to be triggered upon new commits and PRs to the repository.
+
+Artifacts that you build in your tenant namespace (from git pull request, merge request or push events) are pushed to an OCI registry
+along with their supporting metadata (including xref:metadata:index.adoc[SLSA provenance attestations and SBOMs]).
+
 === Application
-An Application CR owns multiple Components. You can think of it as the closest object in {ProductName} that
-models an ongoing supported product version. If for some reason you decide you need to, you can decompose your
-product version further and model different sub-products each as their own Application.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+team. One team can span multiple namespaces if needed, each with many
+applications with components. Namespaces should NOT be shared by many teams.
+
+In your tenant namespace, you will create Components and Applications, run the
+build PipelineRuns that are defined in your git repositories, view and iterate
+on the results of your IntegrationTestScenarios, and create Releases to release
+specific Snapshots according to specific ReleasePlans.
+
+NOTE: You can grant, modify, or remove other users' access to any namespace for
+which you have the `admin` role. Each namespace's admins have the
```
arewm, line 34: @filariow, do you have thoughts for how it might be good to refer to these various roles in the documentation?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+responsibility to ensure that appropriate access is granted to users. If users
+are over-privileged, they might have access to any added secrets or to perform
+other unintended actions.
```
arewm, line 37: Is this something that you want to call out here or should we call it out when talking about adding users? Part of the challenge that we have to contend with is information overflow when people are just getting started.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
 
 ==== Managed namespace
-Sometimes users will need to perform actions that require credentials in someone else's control. Since admins
-in a tenant namespace can see all secrets, these actions need to be performed via separate managed namespaces.
-The primary interaction mode between tenant and managed namespaces is to create a Release referencing a specific
-Snapshot which will trigger a specific pipeline in the managed namespace.
+These are namespaces where a privileged release engineering team or
+site-reliability engineering team manages release pipelines and credentials for
```
arewm, line 41: In https://konflux-ci.dev/docs/releasing/, we don't use these terms but "managed environment team." Are these too close to Red Hat-isms? Can we be consistent here with content elsewhere?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+your organization.
+
+The primary interaction mode between tenant and managed namespaces is to create
+a Release in a tenant namespace referencing a specific Snapshot which will
+trigger a specific release pipeline in the managed namespace.
```
arewm, line 46: Should we link to other topics/headings here? Or link to items in the glossary?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
-its latest built commit, initial build configuration parameters (i.e. Containerfile path), and any relationships to
+its latest built commit, initial build configuration parameters, and any relationships to
 other Components. The CR also contains a reference to its single owning Application.
-Component names must be unique in a namespace, even when components are used in different application.
-
-=== Pipelines as Code (PAC)
-https://pipelinesascode.com/[Pipelines as Code] is a project that provides opinionated tooling to define a
-Tekton-based CI/CD pipeline. It enables the use of a Tekton PipelineDefinition within a git repository and for
-that pipeline to be triggered upon new commits/PR to the repository. In Konflux, these pipelines are pushed to
-the git repositories in the `.tekton` directory.
```
arewm, line 44: Is there motivation for removing this? Is there a better place for it to be moved to?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+Component names are unique in a namespace, even when components are used in different application.
 
 === Build pipeline
-Upon a new push or pull request (merge request in GitLab) event, the pipeline defined in the repository will be
+When you create a Component in {ProductName}, a build pipeline is pushed to the git repository in the `.tekton/`
```
arewm, line 69: Technically, multiple pipelines are pushed. That might be too much of a nitpick/detail to include here though.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+Tasks such as cloning the git repository, prefetching dependencies, building the OCI artifact and source SBOM, and
+generating the source container. The test process includes Tekton Tasks such as running Snyk scans, checking for
 CVEs with clair-in-ci, and running an antivirus scan on the artifact.
 
+{ProductName} inherits the pattern of defining Tekton build pipelines in git
+from its use of https://pipelinesascode.com/[Pipelines as Code (PaC)]. PaC
+enables the use of a Tekton PipelineDefinition within a git repository and for
+that pipeline to be triggered upon new commits and PRs to the repository.
+
+Artifacts that you build in your tenant namespace (from git pull request, merge request or push events) are pushed to an OCI registry
```
arewm, line 83: I don't think we need this parenthetical phrase.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,99 @@ xref:glossary:index.adoc[glossary].
...
+along with their supporting metadata (including xref:metadata:index.adoc[SLSA provenance attestations and SBOMs]).
+
 === Application
-An Application CR owns multiple Components. You can think of it as the closest object in {ProductName} that
-models an ongoing supported product version. If for some reason you decide you need to, you can decompose your
-product version further and model different sub-products each as their own Application.
+An Application CR owns multiple Components. You can think of it as the object in {ProductName} that
+models an ongoing supported product version.
```
arewm, line 88: This feels like a Red Hat ism. Can/should we describe it more generically?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -77,21 +113,21 @@ build is "promoted" to update the references on the Component CR.
...
-contains multiple Components.
+for example, all Components pass the policy requirements, the policy evaluation is true. If a single
+Component in a Snapshot fails the policy, however, the overall result is a failure even if all of the rest have
+clean passes.
 
-NOTE: It is possible to enable Single Component mode when the ECP is being evaluated. See
-xref:patterns:testing-releasing-single-component.adoc[Testing and Release a Single Component]
+NOTE: This behavior can present issues when running the default enterprise contract ITS when an Application
+contains multiple Components. It is possible to enable "Single Component mode" when integration tests or release
+pipelines evaluate the ECP. See xref:patterns:testing-releasing-single-component.adoc[Testing and Release a Single Component]
```
arewm, line 125: I know that we added this here to address a specific challenge. It doesn't quite feel like the right place to be though. Can we instead move this into Conforma sections?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -100,46 +136,22 @@ run and a given ECP which needs to pass for any Snapshot before that pipeline ca
...
-and iterate on the results of the IntegrationTestScenarios, and create Releases for specific Snapshots.
-
-=== What do I do with artifacts built from my namespaces
-
-Artifacts build in your namespace (from git pull/merge request or push events) will be pushed to an OCI registry
-along with their supporting metadata (including xref:metadata:index.adoc[SLSA provenance attestations and SBOMs]).
-Once the artifacts are pushed to the registry, they can be used for any common intended activity including development,
-testing, deployments, and releasing (i.e. pushing the images elsewhere either with credentials that you own or that
-someone else owns). The model of building artifacts in {ProductName} is to create them once and release them to as many
-places as needed/desired where each release action have a unique ECP gating the actions.
```
arewm, line 142: Should we represent this somewhere else? It is an underlying concept that you can release anywhere as long as the policy gating that process succeeds.

arewm (commented):
Thanks for these improvements. I wonder if it makes sense to just try to add this type of content into the glossary as that is essentially what we are doing. 

The issue was originally opened based on a comment from @brunoapimentel (I believe) that getting started content usually has a simple example for how to simply set up an example for exploration.

modules/ROOT/pages/getting-started.adoc
ralphbean: Yeah, it just didn't seem necessary to describe PaC here. It felt like an odd duck in the list.

I moved mention of the .tekton directory to the build pipeline section.

modules/ROOT/pages/getting-started.adoc
ralphbean: I agree, it's a Red Hat ism. Let's just remove it!

modules/ROOT/pages/getting-started.adoc
ralphbean: Not sure - but, let's leave it out for now.

modules/ROOT/pages/getting-started.adoc
ralphbean: but.. not in this PR.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,93 @@ xref:glossary:index.adoc[glossary].
...
+its latest built commit, initial build configuration parameters, and any relationships to
 other Components. The CR also contains a reference to its single owning Application.
-Component names must be unique in a namespace, even when components are used in different application.
-
-=== Pipelines as Code (PAC)
-https://pipelinesascode.com/[Pipelines as Code] is a project that provides opinionated tooling to define a
-Tekton-based CI/CD pipeline. It enables the use of a Tekton PipelineDefinition within a git repository and for
-that pipeline to be triggered upon new commits/PR to the repository. In Konflux, these pipelines are pushed to
-the git repositories in the `.tekton` directory.
+Component names are unique in a namespace, even when components are used in different application.
```
arewm, line 62:
```suggestion
Component names are unique in a namespace, even when components are used in different applications.
```

ralphbean:
> I wonder if it makes sense to just try to add this type of content into the glossary as that is essentially what we are doing.
> The issue was originally opened based on a comment from @brunoapimentel (I believe) that getting started content usually has a simple example for how to simply set up an example for exploration.

I see! I think I agree but I don't think I have the energy for that right now. We could merge this now as an incremental improvement and return to start over on the getting started doc (that really is a wholly different doc), or I'm ok too if you just want to drop this.

modules/ROOT/pages/getting-started.adoc
arewm: That's fine. I liked this addition, but I am biased as its author. I feel like this is one of the uniquely unintuitive parts about Konflux. It isn't just about building artifacts, but it is about building them and then doing something with them according to rules defined _out of your control_. A managed namespace is simply one that you don't control. It _could_ be considered a tenant namespace by some other set of users. But those users are the individuals that define the requirements for passing the namespace barrier successfully.

arewm:
> I think I agree but I don't think I have the energy for that right now. We could merge this now as an incremental improvement and return to start over on the getting started doc (that really is a wholly different doc), or I'm ok too if you just want to drop this.

I would be fine to improve what we have right now. We are not claiming that we are resolving the issue, so it can remain open for further improvements.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,93 @@ xref:glossary:index.adoc[glossary].
...
+you're familiar with common Kubernetes resources such as Pods and Deployments,
+you'll find that {ProductName} resources appear on cluster in the same way:
+Applications, Components, Snapshots, and PipelineRuns.
+
+A consequence of this fact is that commonplace Kubernetes client tools such as
+link:https://kubernetes.io/docs/reference/kubectl/[kubectl] understand and can
+work with {ProductName} APIs.
 
 === Component
 A Component CR describes the properties for an OCI artifact including the git repository where the artifact is from,
```
brunoapimentel, line 59: It might be just me, but I wonder if OCI artifact is a term well-known enough that it does need any further explanation. We could add a tiny explanation in the glossary and link the references there (or link them to an official definition).

modules/ROOT/pages/getting-started.adoc
```diff
@@ -7,63 +7,93 @@ xref:glossary:index.adoc[glossary].
...
-Snapshot which will trigger a specific pipeline in the managed namespace.
+These are namespaces where a managed environment team manages release pipelines
+and credentials for your organization.
+
+The primary interaction mode between tenant and managed namespaces is to create
+a Release in a tenant namespace referencing a specific Snapshot which will
+trigger a specific release pipeline in the managed namespace. See
+xref:releasing:index.adoc[release documentation] for more.
+
+=== Custom Resource (CR)
```
brunoapimentel, line 44: Tiny suggestion: since we're adding this CR section here, we could nest all the following sections under it, since they are all CRs as well (components, pipelines, application, etc). This would make it similar to what we did by adding  "Managed namespace" and "Tenant namespace" under "Namespaces".

brunoapimentel (approved):
As Andrew mentioned, I also imagine the Getting Started section in a completely different way, but this is definitely an improvement compared to what we currently have. I left two small suggestions, feel free to dismiss them if you prefer.

modules/ROOT/pages/getting-started.adoc
ralphbean: Done in 734d676 (I nested "build pipeline" under Component).

modules/ROOT/pages/getting-started.adoc
ralphbean: Updated in 57d76dc.

brunoapimentel (approved):

PR #344, opened 2025-05-16
Add note about removing component right after creation resulting in

rcerven:
orphaned imageRepository and UI failing with imageRepository already exists

### DIFF
--- a/modules/building/pages/creating.adoc
+++ b/modules/building/pages/creating.adoc
@@ -39,6 +39,11 @@ NOTE: For *Gitlab* provider, make sure to xref:building:creating-secrets.adoc#cr
 +
 NOTE: GitHub and GitLab are supported source control providers. GitLab support requires the configuration of xref:building:creating-secrets.adoc#creating-source-control-secrets[source control secrets].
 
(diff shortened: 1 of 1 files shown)


modules/building/pages/creating.adoc
```diff
@@ -39,6 +39,11 @@ NOTE: For *Gitlab* provider, make sure to xref:building:creating-secrets.adoc#cr
 +
 NOTE: GitHub and GitLab are supported source control providers. GitLab support requires the configuration of xref:building:creating-secrets.adoc#creating-source-control-secrets[source control secrets].
 
+NOTE: When Component is immediately removed after creation it might result with orphaned ImageRepository which ownership
+wasn't yet assigned to the Component and when trying to create Component again with the same name UI will
+complain with error that ImageRepository already exists, solution is to remove manually ImageRepository
```
mmorhun, line 44: I think it worth a feature for UI, so it can remove or reuse ImageRepository in such case.

mmorhun (approved):

modules/building/pages/creating.adoc
rcerven: filed https://issues.redhat.com/browse/KFLUXUI-475

PR #345, opened 2025-05-21
Add hidden machine readable content

arewm:
We may want to process documentation to load it into AI models' context windows for improving user troubleshooting. This content might not be terribly useful for humans reading the pages.

This change creates a framework for adding machine-readable content. While the helfulness of this content isn't currenly known, the process show here can be repeated in the future.


ernesgonzalez33:
@arewm I don't exactly understand how this works. Can you please explain it? Maybe we should add instructions also on how to do it in our docs guide

dirgim (approved):
LGTM

arewm:
Ooh, that's a good find. I think that is something similar, yes.

ralphbean:
@arewm I'm going to close this one for now. Re-open if you want to restart discussion on it.

PR #352, opened 2025-06-01
doc(STONEINTG-1206): pipelinerun support for ITS

jsztuka:
Description and guide of newly added parameter **resourceKind** into integration-service ecosystem.
Example of how users can set their integration test scenario to customize parameters for pipelineRun instead of pipeline.

### DIFF
--- a/modules/testing/pages/integration/creating.adoc
+++ b/modules/testing/pages/integration/creating.adoc
@@ -133,6 +133,45 @@ spec:
 
 .. For additional instructions on adding an integration test, see Adding an integration test.
 
+== Customize pipelineRun definition
+
+Integration service provides customization for both pipeline and pipelineRun definitions.
+There are certain attributes that can be defined only for pipelineRuns such as:
+
+* pipeline timeouts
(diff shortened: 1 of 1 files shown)


modules/testing/pages/integration/creating.adoc
```diff
@@ -133,6 +133,45 @@ spec:
...
+  resolverRef:
+    resolver: git
+    resourceKind: pipelinerun
+    params:
+      - name: url
+        value: https://github.com/konflux-ci/integration-examples
+      - name: revision
+        value: main
+      - name: pathInRepo
+        value: pipelines/integration_pipeline_pass.yaml
```
hongweiliu17, line 172: This file `integration_pipeline_pass.yaml` is one pipeline yaml file, can it be used as pipelinrun?

jsztuka: AFAIK, we get the definition of pipeline then encode it in base64. THen just before we generate pipelinerun we decode this base64 string and create new pipelinerun.

PR #354, opened 2025-06-03
Update mintmaker schedule

mkosiarc:
The schedule for dockerfile manager was changed couple of months ago in https://github.com/konflux-ci/mintmaker/commit/cd1498d4603dd65ad61beef923aaa54a05e639f0

It now runs every day after 5am instead of only mondays.


qixiang (approved):
:+1:

PR #358, opened 2025-06-09
fix(KONFLUX-7679): clair scan vulns missing from SCAN_OUTPUT

Josh-Everett:
Clair scan detects blocking high CVE vulnerabilities that are not included in the SCAN_OUTPUT within Konflux. This PR adds a note explaining that only advisories associated with RHSAs will be reported.


dirgim (approved):
LGTM

PR #383, opened 2025-06-16
Remove application.thumbnail annotation

hugares:
This is not used anywhere in konflux so remove this from documentation as it is not needed.

KFLUXSPRT-3653


arewm (approved):
Thanks!

PR #384, opened 2025-06-17
Create space for a partial about gitops releases

ralphbean:
Example of downstream usage: https://gitlab.cee.redhat.com/konflux/docs/users/-/merge_requests/382


ralphbean:
@arewm I updated the partial with a suggestive note.

PR #392, opened 2025-07-01
fix(stoneintg-1243): incl info on roles and rolebindings

jencull:
also include restartPolicy in yaml example of cronJob

### DIFF
--- a/modules/testing/pages/integration/periodic-integration-tests.adoc
+++ b/modules/testing/pages/integration/periodic-integration-tests.adoc
@@ -19,10 +19,18 @@ If the Integration Test is designed to only run periodically rather than during 
 
 === Procedure with example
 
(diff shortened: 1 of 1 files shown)


14rcole:
Since the user asked about a working example, could you put a simple example CronJob in the doc as well?

jencull:
> Since the user asked about a working example, could you put a simple example CronJob in the doc as well?

@14rcole there already is, look at 2 here, just scroll down a bit. Thats the section I added the restartPolicy field
https://konflux.pages.redhat.com/docs/users/testing/integration/periodic-integration-tests.html

Josh-Everett (commented):
LGTM but not sure if running `npm audit fix` will fix the CI error, unfamiliar with that failure.

PR #397, opened 2025-07-10
feat(RELEASE-1493): add user documentation to community-catalog

Paul123111:
Add note about community-catalog to Konflux docs.


modules/releasing/pages/tenant-release-pipelines.adoc
```diff
@@ -175,6 +175,9 @@ spec:
 
 If you write a good reusable release pipeline, please submit it to our link:https://github.com/konflux-ci/release-service-catalog[catalog] so others can use it.
```
johnbieren, line 176: This page is about tenant release pipelines, so I think your new note should be in place of this one, not in addition to. Now that we have community catalog, there is no need to have any tenant/final pipeline resources in release-service-catalog

johnbieren (approved):

PR #399, opened 2025-07-15
Add reminder to RHSM activation keys page

MartinBasti:
Preparing just lockfile using RHSM activation keys is not enough. Pipeline must be also configured to do prefetching of RPMs.


modules/building/pages/activation-keys-subscription.adoc
```diff
@@ -417,3 +417,5 @@ TIP: If you see warnings like "`+WARNING:root:No sources found for...+`" then th
 . Exit the container, and commit the `+rpms.in.yaml+`, `+rpms.lock.yaml+` and `+redhat.repo+` to
 source control. Konflux will use these files to prefetch RPMs for hermetic
 builds.
+
+. Reminder. You still need to xref:ROOT:building:prefetching-dependencies.adoc#rpm[enable prefetching for RPMs] if this step hasn't been done yet.
```
eskultety, line 421: I presume this is because users (apparently) incorrectly assumed that without
```
- default: '{"type": "rpm", "path": "."}'
        description: Build dependencies to be prefetched by Cachi2
        name: prefetch-input
        type: string
```
the pipeline would still automatically process the RPM deps? :)

eskultety (approved):

PR #403, opened 2025-07-17
docs(RELEASE-1352): replace `auto-release` with `block-releases` label

johnbieren:
This commit updates the documentation to deprecate the `auto-release` label on ReleasePlanAdmissions (unchanged on ReleasePlans) with the `block-releases` label. They work the same, but opposite, so auto-release true works the same as block-releases false.


mmalina (approved):

seanconroy2021 (approved):

dirgim (approved):
LGTM

PR #406, opened 2025-07-28
docs: Simplify rapidast docs and link to new exmaple

sfowl:
~Draft for now, as link to the new example won't be ready until this merges:~

https://github.com/RedHatProductSecurity/rapidast/pull/401

EDIT: ready for merge 

Supports [PSSECAUT-1228](https://issues.redhat.com/browse/PSSECAUT-1228)

### DIFF
--- a/modules/testing/pages/integration/third-parties/rapidast.adoc
+++ b/modules/testing/pages/integration/third-parties/rapidast.adoc
@@ -1,59 +1,30 @@
 = RapiDAST
 
-In this guide, you'll learn how to xref:testing:integration/adding.adoc[add a custom integration test] in {ProductName} that uses link:https://github.com/RedHatProductSecurity/rapidast[RapiDAST] as a third-party tool for dynamic application security testing. 
+link:https://github.com/RedHatProductSecurity/rapidast[RapiDAST] is a tool for performing dynamic application security testing (DAST) on _running_ applications.
+
+Therefore, it's recommended to use a xref:testing:integration/adding.adoc[custom integration test] to deploy the target application, and perform a RapiDAST scan inside that same integration test.
+
+NOTE: RapiDAST scans are intended for testing environments and should not be used on production systems.
 
 .Prerequisites
 
-. You have xref:building:creating.adoc[created an application] in {ProductName}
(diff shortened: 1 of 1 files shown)


modules/testing/pages/integration/third-parties/rapidast.adoc
```diff
@@ -1,59 +1,30 @@
...
-
-. You have selected the scanner you want to use with RapiDAST, ZAP, Nessus or a generic scanner. Refer to the RapiDAST link:https://github.com/redHatProductSecurity/rapidast/?tab=readme-ov-file#scanners[documentation] for available scanners and their configurations.
-
-. You have access to a git repository containing the link:https://github.com/RedHatProductSecurity/rapidast?tab=readme-ov-file#an-openapi-schema[OpenAPI] specification of your application or another supported entry point (e.g., endpoints, spiders, etc.).
+Therefore, it's recommended to use a xref:testing:integration/adding.adoc[custom integration test] to deploy the target application, and perform a RapiDAST scan inside that same integration test.
 
-. Optional - You have access to link:https://github.com/RedHatProductSecurity/rapidast?tab=readme-ov-file#exporting-to-google-cloud-storage[Google Cloud Storage] to facilitate export and storage of scan results.
```
hongweiliu17, line 17: Are these info not needed any more?

sfowl: It's a bit complicated, this is just another option in a rapidast config file so it's possible and optional to use it, just like any other option in the rapidast docs.

That said, neither the original example, nor newer example contain a full demo of how to embed any secrets in a rapidast config file (which is required for this feature), it's something I'd like to improve later.

Also, the option of pushing scan results to a google bucket sort of competes with the model of pushing results to quay.io (like used for sast tasks). We have a [separate ticket](https://issues.redhat.com/browse/PSSECAUT-1230) for that, and until we reach consensus I don't want cause confusion for users on which option they should use.

So for now I think it's best to drop explicit mention of this feature on this page.

hongweiliu17: It loos good to me. Thanks!

hongweiliu17 (approved):
LGTM

PR #419, opened 2025-08-13
docs(ISV-5874): Add a reference to Mobster SBOM generator

Allda:
A Mobster is a new tool responsible for SBOM management in Konflux. This commits adds a reference to Mobster documentation where we describe SBOM structure, workflows and content type.

JIRA: ISV-5874


Allda:
@ralphbean @arewm Could you please review this PR? Thanks!

PR #421, opened 2025-08-14
docs: Add integration testing guide

Troy876:
Document on integration testing best practices and links to tasks/stepactions that will point to tekton-integration-catalog with a short description.

[KFLUXDP-231](https://issues.redhat.com/browse/KFLUXDP-231)

### DIFF
--- a/modules/testing/pages/integration/integration-testing-guide.adoc
+++ b/modules/testing/pages/integration/integration-testing-guide.adoc
@@ -1,0 +1,173 @@
+== Best Practices for Integration Testing
+
+=== Design Atomic Tests
+
+* **Single Responsibility**: Each integration test should focus on testing one specific integration point or workflow
+* **Isolation**: Tests should be independent and not rely on the state from other tests
+* **Deterministic**: Tests should produce consistent results across different environments
+
+=== Leverage Konflux's Snapshot-based Testing
+
+* **Snapshot Composition**: Konflux automatically creates snapshots combining the latest tested components with your new build
+* **Atomic Updates**: Only one component reference is updated per snapshot, ensuring each artifact can be tested independently
+* **Integration Test Scenarios**: Use `IntegrationTestScenario` resources to define your test suites
+
+== Available Tasks and StepActions from Konflux Repositories
+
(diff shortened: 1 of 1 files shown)


ascerra:
> Is this just a page documenting the various tasks in https://github.com/konflux-ci/tekton-integration-catalog/tree/main? If so, I think the documentation should exist on _that_ page. We can add a single link out to the repository to indicate that it is there. If we have an index/summary of tasks here, however, it is much more likely to be out of date. The test/helper tasks should all have good documentation and examples in the repository so that users can easily get the contextual relevance.
> 
> Some of this content may be useful but it doesn't need a separate page. The content that we want to keep here can just be merged into the https://konflux-ci.dev/docs/testing/, https://konflux-ci.dev/docs/testing/build/, and/or https://konflux-ci.dev/docs/testing/integration/ pages.

@arewm we had the same concerns about duplicating this documentation, one of the leading reasons we wanted to make it available within the Konflux docs themselves is that those docs are given to users by the bot in the Konflux-users channel when they have issues and we believe some of these tasks will help them with their issues.

We'd thought of finding a way to auto update the konflux docs when the tekton-integration-catalog changes but now I'm curious if our efforts would be better spent trying to get the support bot trained on the docs in the tekton-integration-catalog repo.

arewm:
> one of the leading reasons we wanted to make it available within the Konflux docs themselves is that those docs are given to users by the bot

> We'd thought of finding a way to auto update the konflux docs when the tekton-integration-catalog changes but now I'm curious if our efforts would be better spent trying to get the support bot trained on the docs in the tekton-integration-catalog repo.

I think that this is the thing to do... ingest that repo into the bot's context.

ascerra:
Alright nice so the support bot is already ingesting the tekton-integration-catalog repo ([See slack thread](https://redhat-internal.slack.com/archives/C073CAS5YUQ/p1756306570216129))

So this just leaves me with the question of where we can provide a link to this repo in the konflux user docs to help someone who is reading docs but hasn't asked a question in the support channel yet

arewm (approved):

PR #437, opened 2025-08-29
Add --platform to cosign CLI examples

### LINKED ISSUE #435: add --platform information to cosign CLI example
The https://konflux-ci.dev/docs/metadata/sboms/ page (`modules/metadata/pages/sboms.adoc`) has CLI guidance for the `cosign` command, like so:

```
$ cosign download sbom $IMAGE
```

If the image is multi-arch, then this command will give the sbom for the *image index*. The clue is that it will show a couple of references to `"SPDXRef-image-index"`. This command will not list any traditional software packages (RPMs, Python wheels, etc).

To list packages, the user must specify a `--platform` argument, for example, `linux/amd64` or `linux/arm64`:

```
cosign download sbom --platform linux/amd64 quay.io/example-workloads/example-tenant/my-container-1-0/my-container-1-0:latest
```

ktdreyer:
Updates the cosign CLI examples in the SBOM documentation to include the `--platform linux/amd64` option. Without this option, multi-arch images return only the image index SBOM instead of actual software packages (RPMs, Python wheels, etc).

Fixes #435

### DIFF
--- a/modules/metadata/pages/sboms.adoc
+++ b/modules/metadata/pages/sboms.adoc
@@ -54,7 +54,7 @@ In the CLI, complete the following steps to download the SBOM for a component:
 +
 [source]
 ----
-$ cosign download sbom $IMAGE
+$ cosign download sbom --platform linux/amd64 $IMAGE
 ----
 
 +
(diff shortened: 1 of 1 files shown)


ralphbean:
Awesome - thanks @ktdreyer!

PR #440, opened 2025-09-02
Update doc links to Conforma

Acepresso:
Update references in the Konflux docs to Conforma following recent rename of Conforma repositories and reorg in the docs.

Ref: https://issues.redhat.com/browse/EC-1456


ralphbean (approved):
Thank you!

PR #447, opened 2025-09-04
Update CODEOWNERS to include ben-alkov for prefetch

ben-alkov:
Hermeto team needs to be notified when changes are made to this file, so we can ensure that it stays accurate.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -7,6 +7,7 @@
 
 # build
 /modules/building/	@konflux-ci/build-maintainers
+/modules/building/pages/prefetching-dependencies.adoc  @ben-alkov
 
 # integration
 /modules/testing/       @konflux-ci/integration-service-maintainers


CODEOWNERS
```diff
@@ -7,6 +7,7 @@
 
 # build
 /modules/building/	@konflux-ci/build-maintainers
+/modules/building/pages/prefetching-dependencies.adoc  @ben-alkov
```
MartinBasti, line 10: Having only one memeber in codeowners doesn't scale, it will require all the review from you only and block merging. I'd suggest to create a hermeto team and put the whole team here

MartinBasti: we can reuse hermeto team also in build definitions

eskultety: Can we simply use `build-maintainers` for the time being?

ben-alkov: If we do a larger group, I'll need to add in '.*rpm.*md' as well.

Q: Who is in `build-maintainers`?

eskultety: > adding hermeto ppl into build maintainers?
> -    you will be notified about many things you may not care about

True.

> adding build-maintainers into /modules/building/pages/prefetching-dependencies.adoc
> it's already there for the parent dir, so it should be inherited

True, didn't notice it in the diff. This is still a draft and I don't think this is a critical change, so maybe we can setup a prefetch subteam in the GH org and then refresh this PR accordingly?

ben-alkov: > so maybe we can setup a prefetch subteam in the GH org and then refresh this PR accordingly?

Yeah, this was Martin's idea of setting up a net-new Hermeto team for this, and also build-definitions.

PR #456, opened 2025-09-15
Add script to centralize pipeline def

lcarva:


lcarva:
> This looks good (I haven't tested it though). Do you think it still makes sense to have the other documentation/scripts in addition to this?

I think it helps understand the process a little more. I see the script as a "cheat code" to make it procedure faster.

ralphbean (approved):

MartinBasti (approved):

PR #471, opened 2025-09-30
fix cosign installation URL

ktdreyer:
The upstream cosign devs have moved the install instructions to a new
location. Update the link to reflect this change.


MartinBasti (approved):
Thank you!

PR #475, opened 2025-10-01
Adding permissions and roles table that was in an ADR, and also adding short procedure on configuring roles.

Paul2024-ship-it:
This PR takes the existing ADR  on permissions: [https://konflux-ci.dev/architecture/ADR/0011-roles-and-permissions.html](url) and converts it for the upstream docs. I have also added a brief section on using the yaml files for configuration. Note this section lists four roles.


Paul2024-ship-it:
Discussion ongoing as to what roles need to be documented.

PR #484, opened 2025-10-14
Provide clarification on Global Candidate List

Josh-Everett:
Update notes about integration service's GCL implementation to provide better context for users.


hongweiliu17 (approved):
LGTM

PR #487, opened 2025-10-20
feat(STONEINTG-1370): update kube-api docs

jsztuka:
Integration-service updated their integrationTestScenario api documentation, in order to reflect it in konflux.docs we need to run `api-gen` to update those values. Together with integration service, other services were updated as `api-gen` fetches api objects from various places.

### DIFF
--- a/modules/reference/pages/kube-apis/application-api.adoc
+++ b/modules/reference/pages/kube-apis/application-api.adoc
@@ -30,7 +30,7 @@ Package v1alpha1 contains API Schema definitions for the appstudio v1alpha1 API 
 
 
 
-Application is the Schema for the applications API.  For a detailed description with examples, refer to <a href="https://github.com/redhat-appstudio/book/blob/main/book/HAS/hybrid-application-service-api.md"> Hybrid Application Service Kube API </a>
+Application is the Schema for the applications API. For description, refer to xref:kube-apis/application-api.adoc[Application API Reference] Hybrid Application Service Kube API </a>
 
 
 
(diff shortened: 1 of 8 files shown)


modules/reference/pages/kube-apis/application-api.adoc
```diff
@@ -204,7 +204,7 @@ ComponentDetectionDescription holds all the information about the component bein
 | *`language`* __string__ | Language specifies the language of the component detected +
 Example: JavaScript + |  | 
 | *`projectType`* __string__ | ProjectType specifies the type of project for the component detected +
-Example Node.JS + |  | 
+Example Node. JS + |  |
```
MartinBasti: Why whitespace was injected?

modules/reference/pages/kube-apis/conforma.adoc
```diff
@@ -0,0 +1,279 @@
+// Generated documentation. Please do not edit.
```
MartinBasti: This is new doc, it shoule be added into nav page https://github.com/konflux-ci/docs/blob/main/modules/reference/nav.adoc

However I see there enterprise contract page, should be enterprise contract page removed/replaced by conforma?

jsztuka: Good point, I would change the nav.doc so it points to the conforma honestly. Let me change it.

jsztuka: Agreed, removed.

PR #496, opened 2025-10-24
Mention how users can request a new manager support

staticf0x:


staticf0x:
Good point, I'll put in the dowstream docs then

PR #505, opened 2025-11-05
remove namespace as it seems to be required

kannon92:
I followed these docs and removed namespace.

The gitlab CI yelled at me because namespace was not specified. I think namespace is required.

### DIFF
--- a/modules/patterns/pages/managing-multiple-versions.adoc
+++ b/modules/patterns/pages/managing-multiple-versions.adoc
@@ -137,7 +137,7 @@ The following kinds of resources are currently supported. Please do not attempt 
 
 Here are specific examples for how to clean up and use the YAML for certain resource kinds:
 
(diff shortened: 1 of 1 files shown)


kannon92:
```
    ) -> None:
        """Verify the tenants doesn't try to deploy to other namespaces"""
        tenant_dir, resource = all_resources
        metadata = resource.data.get("metadata", {})
        namespace = metadata.get("namespace")
        kind = resource.data.get("kind")
        if kind == "Namespace":
            namespace = metadata["name"]
    
        if not namespace:
>           pytest.fail("Namespace was not defined in the manifest")
E           Failed: Namespace was not defined in the manifest
```

If I tried to follow the docs and remove namespace from my ProjectDevelopmentStream

MartinBasti:
however docs says, to remove namespace in resources section

PR #522, opened 2025-11-24
Manually update task bundles using pmt

tkdchen:
STONEBLD-3592

The pmt-migrate command now handles both bundle updates and task migrations when available. This functionality is implemented for STONEBLD-3592.

### DIFF
--- a/modules/troubleshooting/pages/builds.adoc
+++ b/modules/troubleshooting/pages/builds.adoc
@@ -88,11 +88,19 @@ cat <<'EOF' > update-tekton-task-bundles.sh
 #!/bin/bash
 
 # Use this script to update the Tekton Task Bundle references used in a Pipeline or a PipelineRun.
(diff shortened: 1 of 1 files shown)


modules/troubleshooting/pages/builds.adoc
```diff
@@ -121,7 +121,7 @@ for old_ref in ${OLD_REFS}; do
     [[ $new_ref == $old_ref ]] && continue
     echo "New digest found! $new_ref"
     for file in $FILES; do
-        sed -i -e "s!${old_ref}!${new_ref}!g" $file
+        pmt migrate --new-bundle "$new_ref" --pipeline-file "$file" <1>
     done
```
chmeliik: Would it be more efficient if we ran `pmt migrate` on all the files at the same time, instead of doing it one by one? I imagine we would only have to do one round of migrations resolution in that case, which could make a significant difference for repos with many pipelines

PR #523, opened 2025-11-24
initial pulp doc

YasenT:
A doc describing the pulp-access-controller

### DIFF
--- a/modules/building/pages/pulp-access.adoc
+++ b/modules/building/pages/pulp-access.adoc
@@ -64,55 +64,9 @@ Both conventions are functionally equivalent.
 
 == Using it in builds
 
-After the secret is created, mount it in your build pods. The following example demonstrates using it in an RPM build pipeline:
+After the secret is created, mount it in your build pods.
 
-[source,yaml]
-----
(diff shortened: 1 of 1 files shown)


modules/building/pages/pulp-access.adoc
```diff
@@ -0,0 +1,179 @@
...
+Both work the same way.
+
+== Using it in builds
+
+Once the secret is created, you can mount it in your build pods. Here's an example of using it in an RPM build pipeline:
+
+[source,yaml]
+----
+- name: rpmbuild-and-push
+  image: quay.io/fedora/fedora:latest
```
MartinBasti: Neither `pulp` nor `rpmbuild` are part of the image, so this example doesn't work.

I'd expect some better integration with RPM build task, not that users must write custom rpmbuild and push task. Will be custom build task even accepted by conforma?

YasenT: I meant it as a high-level type of example, configure your pulp using the cert etc, and make use of it. And I wanted to make it simple
But maybe it shouldn't be there at all?

PR #592, opened 2026-03-11
OLS-2316 - Adding section about reporting periodic tests

JoaoFula:


MartinBasti (changes requested):
I don't think that this belongs to upstream documentation.

What is the purpose of this for common konflux user? This seems like very specific use case not a general one.

Examples are full of Red Hat specific deployments, that's shouldn't be in doc.

Also, dumping 500 lines script as example doesn't look well. If this is something konflux users desire, then it should be a real well tested task bundle, not a example to be fixed.

JoaoFula:
@dirgim https://github.com/konflux-ci/tekton-integration-catalog/pull/279 created and tested. We can adapt the docs to mention this pipeline when merged or ignore the docs changes altogether. Up to you

PR #614, opened 2026-05-07
fix(remote pipelines): improve the remote pipelines recommendation

aThorp96:
Including the pipelines and tasks in the `.tekton` directory reduces API requests via PaC's built-in resolver

PaC's resolver "renders" the pipelineruns before creation by resolving pipelineRefs which don't use other resolvers) and replacing your PLR's pipelineRef with a pipelineSpec. If the pipeline referenced by also exists in the .tekton directory, PaC already read the file when scanning for pipelineRuns so it's already in memory and no additional API requests are made.

### DIFF
--- a/modules/patterns/pages/keep-remote-pipelines-up-to-date.adoc
+++ b/modules/patterns/pages/keep-remote-pipelines-up-to-date.adoc
@@ -21,7 +21,7 @@ If the repository is private, be sure to have a Secret in your {ProductName} ten
 
 .*Procedures*
 
-. Add Tekton Pipelines and Tasks in the repository. A suggested repository structure would be:
+. Add Tekton Pipelines and Tasks in the repository. Here is a suggested repository structure:
 
 +
 [source,shell]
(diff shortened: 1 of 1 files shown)


modules/patterns/pages/keep-remote-pipelines-up-to-date.adoc
```diff
@@ -21,21 +21,21 @@ If the repository is private, be sure to have a Secret in your {ProductName} ten
 
 .*Procedures*
 
-. Add Tekton Pipelines and Tasks in the repository. A suggested repository structure would be:
+. Add Tekton Pipelines and Tasks in the repository. Here is a suggested repository structure:
```
aThorp96, line 24: Running the CI locally the previous wording triggered the rule `Konflux.ConsistentTense` due to the "would be"

arewm (approved):
Thanks for this. Do you want to be clearer about the anti-pattern, potentially using a caution or warning (I am not sure the exact one to use) call-out?

lcarva (approved):

aThorp96:
Done @arewm

arewm (approved):

PR #617, opened 2026-05-18
doc(STONEINTG-1434): testGraph logic

jsztuka:
User oriented documentation on how to specify exact order of integration test scenarios within their component group.

Three different examples were added:

- Parallel tests
- Sequence tests
- Order tests

### DIFF
--- a/modules/testing/pages/integration/order.adoc
+++ b/modules/testing/pages/integration/order.adoc
@@ -1,0 +1,145 @@
+= Specifying order of integration tests
+:description: Use the component group model in {ProductName} to define the order of integration test scenarios. Specify the exact order of execution for your integration tests.
+
+Refer to the following examples for inspiration. These examples assume you have multiple integration tests created.
+
+TestGraph is the struct (*map[string][]TestGraphNode*) that holds such information. Each node of the TestGraph must contain Name of the integration test scenario and have optional attribute *failFast*.
+
+*failFast* describes how to behave if integration test fails.
+
+.Options: 
+*"false"* (default) - continues running dependent tests, *"true"* - skips dependent tests
+
+[IMPORTANT]
+====
+Your integration tests are not required to be contained within testGraph by default. You need to specify them under testGraph only in case you want to run them in specific order.
+
+====
+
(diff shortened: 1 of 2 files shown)


modules/testing/pages/integration/order.adoc
```diff
@@ -0,0 +1,136 @@
...
+      - name: clamav-scan
+      - name: dast-tests
+        failFast: false
+      - name: operator-scorecard
+    e2e-test:
+      - name: clamav-scan
+        failFast: true
+    operator-scorecard:
+      - name: deployment
+        failFast: false
```
14rcole, line 135: I think we should show that if _some_ tests are order-dependent but others are not, those tests will run independently.  Can_ you expand example 2 or 3 to show or explain that?

modules/testing/pages/integration/order.adoc
```diff
@@ -0,0 +1,136 @@
...
+    second-step:
+      - name: dast-test
+        failFast: true
+    first-step:
+      - name: clamav-scan
+---
+
+This is the order of integration tests execution according to the testGraph definition above:
+
+. clamav-scan - in case clamav-scan fails, second-step follows execution with the rest of the chain.
```
14rcole: Can you add another test that is not part of the test chain? Basically say "this ITS belongs to the ComponentGroup but isn't in the testGraph. Because of that, we run it alongside clamav-scan as if it has no parents"

dirgim (approved):
LGTM, agree with Ryan's suggestions

modules/testing/pages/integration/order.adoc
jsztuka: Added, let me know if that works.

modules/testing/pages/integration/order.adoc
jsztuka: Added, let me know if that works.

jsztuka:
/on-hold
This PR will be merged once https://redhat.atlassian.net/browse/STONEINTG-1281 gets implemented.

14rcole (approved):

PR #624, opened 2026-06-26
move periodic-integration-tests rbac to partial

filariow:
We want to be able to overwrite the section on how to configure RBAC so that customers can customize that as per their needs.


dirgim (approved):
LGTM

PR #641, opened 2026-08-06
STONEBLD-4926 docs: add task documentation page generated from build-pipeline-tasks

yash2189:
## Summary
- Add a Task documentation page under Building → Build configuration that lists the latest version of each task from [container-build-catalog/task](https://github.com/konflux-ci/container-build-catalog), with links to their READMEs
- Add `hack/gen-task-docs.sh` and `npm run task-gen` to regenerate the page

Closes [STONEBLD-4926](https://redhat.atlassian.net/browse/STONEBLD-4926)

[STONEBLD-4926]: https://redhat.atlassian.net/browse/STONEBLD-4926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hack/gen-build-tasks-docs.sh
+++ b/hack/gen-build-tasks-docs.sh
@@ -1,0 +1,42 @@
+#! /bin/bash
+
+set -e
+
+repo_root="$(pwd)"
+dest="${repo_root}/modules/building/pages/task-documentation.adoc"
+repo="container-build-catalog"
+genSrc="$(mktemp -d /tmp/container-build-catalog.XXXXXX)"
+trap 'rm -rf "${genSrc}"' EXIT
+
+pushd "${genSrc}"
+
+git clone --depth 1 "https://github.com/konflux-ci/${repo}.git" "${repo}"
+
+cat > "${dest}" <<'EOF'
+= Task documentation
+:description: Reference of Tekton tasks from the konflux-ci/container-build-catalog repository.
+
+// Generated documentation. Please do not edit.
+// Run: npm run gen-build-tasks-docs
+
+This page lists Tekton tasks defined in
+link:https://github.com/konflux-ci/container-build-catalog/tree/main/task[container-build-catalog/task].
+
+[cols="2,3", options="header"]
+|===
+| Task | Documentation
+EOF
+
+# One row per task that has a README (layout: task/<name>/README.md)
+for readme in "${repo}"/task/*/README.md; do
+    [ -f "${readme}" ] || continue
+
+    name="$(basename "$(dirname "${readme}")")"
+    url="https://github.com/konflux-ci/container-build-catalog/blob/main/task/${name}/README.md"
+    echo "| \`${name}\` | link:${url}[Task README]" >> "${dest}"
+done
+
+echo "|===" >> "${dest}"
+
+popd
+echo "Wrote ${dest}"
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "api-gen": "./hack/gen-api-docs.sh",
+    "gen-build-tasks-docs": "./hack/gen-build-tasks-docs.sh",
     "build": "antora generate --clean --fetch antora-playbook.yml",
     "generate-llms-txt": "bash hack/generate-llms-txt.sh public/llms.txt",
     "preview:netlify": "netlify deploy --dir public",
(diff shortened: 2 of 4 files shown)


hack/gen-task-docs.sh
```diff
@@ -0,0 +1,51 @@
+#! /bin/bash
+
+set -e
+
+genSrc="$(pwd)/task-docs-gen/src"
+dest="$(pwd)/modules/building/pages/task-documentation.adoc"
+repo="build-definitions"
```
slimreaper35: Shouldn't we get "data" from https://github.com/konflux-ci/build-pipeline-tasks instead ?

yash2189: done

hack/gen-task-docs.sh
```diff
@@ -0,0 +1,42 @@
+#! /bin/bash
+
+set -e
+
+genSrc="$(pwd)/task-docs-gen/src"
```
slimreaper35: Wouldn't be better to create a random temporary directory instead like `/tmp/build-pipeline-task.XXXXX` ?

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "api-gen": "./hack/gen-api-docs.sh",
+    "task-gen": "./hack/gen-task-docs.sh",
```
slimreaper35:
```suggestion
    "task-gen": "./hack/gen-build-tasks-docs.sh",
```

naming nitpick, I would emphasize that the script generates build tasks only

.github/workflows/task-documentation.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+'on':
+  schedule:
+    - cron: '0 6 * * 5' # Friday 06:00 UTC
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'hack/gen-task-docs.sh'
+      - 'modules/building/pages/task-documentation.adoc'
+      - '.github/workflows/task-documentation.yaml'
```
slimreaper35, line 28: I don't understand the purpose of this workflow. The docs page just lists the build pipeline tasks. I don't think we will add new one so ofter that we need a workflow for that.

slimreaper35 (commented):
Could you add more information in the commit message as well?

.github/workflows/task-documentation.yaml
yash2189: The workflow is mainly to avoid manual regeneration when task READMEs change since it would refresh the table and open a PR so we don’t have to remember to run npm run task-gen. Happy to drop the Friday schedule and keep manual-only (workflow_dispatch), or remove the workflow entirely and rely on regenerating when we touch tasks. What’s your preference?

hack/gen-task-docs.sh
yash2189: done

.github/workflows/task-documentation.yaml
slimreaper35: My preference would be to drop the workflow completely. The value is marginal. It takes ~1 minute to run the command and open a PR.

yash2189: done

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "api-gen": "./hack/gen-api-docs.sh",
+    "task-gen": "./hack/gen-build-tasks-docs.sh",
```
slimreaper35: nitpick: The script name doesn't match the file.

slimreaper35 (approved):

package.json
yash2189: resolved

slimreaper35 (approved):

hack/gen-build-tasks-docs.sh
```diff
@@ -0,0 +1,42 @@
+#! /bin/bash
+
+set -e
+
+repo_root="$(pwd)"
+dest="${repo_root}/modules/building/pages/task-documentation.adoc"
+repo="build-pipeline-tasks"
```
chmeliik: We'll need to rename to `container-build-catalog` everywhere now that the repo has been renamed

yash2189: done

chmeliik (approved):

PR #646, opened 2026-08-20
docs: document PIP_INDEX_URL support for pip prefetch

LalatenduMohanty:
Add documentation for the new pip-index-url task parameter on the prefetch-dependencies task, which sets PIP_INDEX_URL in the Hermeto container as a pipeline-level alternative to hardcoding --index-url in requirements.txt files.


chmeliik (approved):
LGTM, thanks for the update!

